PR #5, opened 2022-02-22
added configmaps for search deployment

xrajesh:


SherinV (approved):

controllers/common.go
```diff
@@ -0,0 +1,10 @@
+// Copyright Contributors to the Open Cluster Management project
+package controllers
+
+func getLabels(val string) map[string]string {
```
jlpadilla, line 4: Can you add a comment describing this function?  Maybe a better name is `generateLabels()`?

controllers/create_indexerconfigmap.go
```diff
@@ -0,0 +1,65 @@
...
+
+	ns := instance.GetNamespace()
+	cm := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "search-indexer",
+			Namespace: ns,
+		},
+	}
+	data := map[string]string{}
+	data["service"] = ns + "/search-aggregator"
```
jlpadilla, line 51: This should be `/search-indexer`.
This is the name of [this service](https://github.com/stolostron/search-v2-operator/blob/main/controllers/create_indexerservice.go#L47).

controllers/create_indexerconfigmap.go
```diff
@@ -0,0 +1,65 @@
...
+	return nil, nil
+}
+
+func (r *OCMSearchReconciler) IndexerConfigmap(instance *cachev1.OCMSearch) *corev1.ConfigMap {
+
+	ns := instance.GetNamespace()
+	cm := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "search-indexer",
+			Namespace: ns,
```
jlpadilla, line 47: This configmap must have this label.
```
config: acm-proxyserver
```

controllers/create_indexerconfigmap.go
```diff
@@ -0,0 +1,65 @@
...
+	cachev1 "github.com/stolostron/search-v2-operator/api/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+func (r *OCMSearchReconciler) createIndexerConfigmap(request reconcile.Request,
```
jlpadilla, line 16: This function is missing the case where the configmap already exists.  Could we check that the current contents match the expected content?  If it doesn't match, should we change update it or just report the error?

I see this same pattern in several files, maybe we should make it a common function?

controllers/create_indexerconfigmap.go
```diff
@@ -0,0 +1,65 @@
...
+	data["path"] = "/aggregator/clusters/"
+	data["sub-resource"] = "/sync"
+	data["use-id"] = "true"
+	data["secret"] = ns + "/search-indexer-certs"
+	data["caConfigMap"] = "search-ca-crt"
+	cm.Data = data
+
+	err := controllerutil.SetControllerReference(instance, cm, r.Scheme)
+	if err != nil {
+		log.V(2).Info("Could not set control for search-indexer configmap")
```
jlpadilla, line 62: - Should this be a warning?
- Can we include the original err in the log message?

controllers/create_indexerconfigmap.go
xrajesh: Yes - thats coming

jlpadilla (approved):
/lgtm
/hold

PR #11, opened 2022-03-09
Updated Search CRD

xrajesh:


api/v1alpha1/search_types.go
```diff
@@ -0,0 +1,199 @@
+/*
+Copyright 2022.
```
jlpadilla, line 2:
```
// Copyright Contributors to the Open Cluster Management project
```

api/v1alpha1/search_types.go
```diff
@@ -0,0 +1,199 @@
...
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// +optional
+	// Storgeclass Name to be used by default db
+	DBStorage StorageSpec `json:"dbStorage,omitempty"`
+
+	// +optional
+	//Configmap name contains parameters to override default db parameters
+	DbConfig string `json:"dbConfig,omitempty"`
```
jlpadilla, line 39: Does this references a ConfigMap? Referencing another object adds complexity. Why don't we define the required data here? 
Maybe the name could be `ExternalDBInstance`.

api/v1alpha1/search_types.go
```diff
@@ -0,0 +1,199 @@
...
+	AvailabilityConfig AvailabilityType `json:"availabilityConfig,omitempty"`
+
+	// +optional
+	// Control list of Kubernetes resources indexed by search-collector
+	AllowDenyResources FilterSpec `json:"allowDenyResources,omitempty"`
+
+	// +optional
+	// Kubernetes secret name containing user provided db secret
+	// Secret contains connection url, certificates
+	CustomDbConfig string `json:"customDbConfig,omitempty"`
```
jlpadilla, line 56: Maybe the name could be `ExternalDBInstance`.
What's the expected format for the secret?
Typically certs are managed by a cert service (ie. certmanager) so adding additional fields might be difficult.

api/v1alpha1/search_types.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+type ResourceListSpec struct {
+	//API Group names to be filtered
+	APIGroups []string `json:"apiGroups,omitempty"`
+	//Resource names to be filtered
+	Resources []string `json:"resources,omitempty"`
+	// +optional
+	//Cluster Labels this filter to be applied
+	ClusterLabels []string `json:"clusterLabels,omitempty"`
```
jlpadilla, line 134: 1. Need to confirm that we want to use labels vs ClusterSets.
2. If using labels, we should implement as a [LabelSelector](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L1178-L1191) for consistency with K8s. It also adds more flexibility.

api/v1alpha1/search_types.go
```diff
@@ -0,0 +1,199 @@
...
+	// +optional
+	// Denied resources from collector
+	DeniedResources ResourceListSpec `json:"deniedResources,omitempty"`
+}
+
+type ResourceListSpec struct {
+	//API Group names to be filtered
+	APIGroups []string `json:"apiGroups,omitempty"`
+	//Resource names to be filtered
+	Resources []string `json:"resources,omitempty"`
```
jlpadilla, line 131: This needs to be a list of []ApiGroups and []Resources.
Example:
```yaml
  AllowedResources: 
  - apiGroups: [""]
    resources: ["pods", "alertmanager"]
  - apiGroups: ["open-cluster-management.io", "tuned.openshift.io"]
    resources: ["*"]
  DeniedResources:
  - apiGroups: [""]
    resources: ["secrets"]
  - apiGroups: ["open-cluster-management.io"]
    resources: ["policies", "iampolicies", "certificatepolicies"]
```

api/v1alpha1/search_types.go
```diff
@@ -0,0 +1,199 @@
...
+	//Cluster Labels this filter to be applied
+	ClusterLabels []string `json:"clusterLabels,omitempty"`
+}
+
+// SearchStatus defines the observed state of Search
+type SearchStatus struct {
+	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+	// Human readable health state
+	SearchHealth string `json:"searchHealth,omitempty"`
```
jlpadilla, line 142: `Search` is redundant, I'd name this just `Health`.

api/v1alpha1/search_types.go
```diff
@@ -0,0 +1,199 @@
...
+	// Database used by search
+	DBInUse string `json:"dbInUse,omitempty"`
+	// Storage used by database
+	StorageInUse string `json:"storageInUse,omitempty"`
+	// +optional
```
jlpadilla, line 147: Maybe `InUse` is implied and not needed because status represents current state.

api/v1alpha1/search_types.go
```diff
@@ -0,0 +1,199 @@
...
+	Status corev1.ConditionStatus `json:"status"`
+	// Last time the condition transitioned
+	LastTransitionTime metav1.Time `json:"lastTransitionTime"`
+	// +optional
+	// Reason for the condition's last transition.
+	Reason string `json:"reason,omitempty"`
+	// +optional
+	// Human readable message
+	Message string `json:"message,omitempty" `
+}
```
jlpadilla, line 162: Should use the generic `Condition` from the Kubernetes API? https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Condition

api/v1alpha1/search_types.go
```diff
@@ -0,0 +1,199 @@
...
+	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+	// Human readable health state
+	SearchHealth string `json:"searchHealth,omitempty"`
+	// Database used by search
+	DBInUse string `json:"dbInUse,omitempty"`
+	// Storage used by database
+	StorageInUse string `json:"storageInUse,omitempty"`
+	// +optional
+	Conditions SearchConditions `json:"conditions,omitempty"`
```
jlpadilla, line 148: Is it possible to define the array inline like this
```go
Conditions []SearchCondition `json:"conditions,omitempty"`
```

bundle/manifests/search-v2-operator.clusterserviceversion.yaml
```diff
@@ -137,7 +137,8 @@ spec:
                 - --leader-elect
                 command:
                 - /manager
-                image: quay.io/stolostron/search-v2-operator:latest
+                #image: quay.io/stolostron/search-v2-operator:latest
+                image: quay.io/xrajesh/search-v2-operator:a0309
```
jlpadilla, line 141: Revert before merge.

api/v1alpha1/search_types.go
```diff
@@ -0,0 +1,199 @@
...
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// SearchSpec defines the desired state of Search
+type SearchSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// +optional
```
jlpadilla, line 33: @dockerymick Could you help review the code comments in this file? These comments become the API documentation.

dockerymick: Yes! Will review shortly

api/v1alpha1/search_types.go
xrajesh: This is name of the Configmap , isnt it like name of the storageClass @jlpadilla

api/v1alpha1/search_types.go
xrajesh: This secret can like following 
```
data:
        db_url:
        db_name: 
        db_user:
        db_password:
        ca_cert:
```
For example the db can be in a VM , the certs will be provided by the customer right. @jlpadilla

jlpadilla: - If postgres is on a VM, yes, but most likely there will be some automation to move the certs if these need to be rotated periodically.
- I'm thinking of the scenario where postgres is on the cluster, for example, ACM installs a postgres instance for the product and search uses the ACM postgres instead of our own.

Users will need to know the structure of this secret. Can we add something like this?
```
// ... the secret must contain these keys [db_host, db_port, db_user, db_password, db_name, ca_cert]
```

`db_url` is a compressed format like: `postgresql://{db_user}:{db_password}@{db_host}:{db_port}/{db_name}`

api/v1alpha1/search_types.go
jlpadilla: What's the concern with upgrade?  If the kube API changes it will need to be published as `v2`, so it won't break us.

My concern is consistency. This is a concept that kubernetes has standardized across resources, so I don't want us to drift.

api/v1alpha1/search_types.go
xrajesh: Yes Options 2 - is what we want - But creating a configmap from Key,value pair is easy - But this comes in only when we have to tune the DB.

api/v1alpha1/search_types.go
xrajesh: This is mainly for ExternalDB - If ACM has a common Postgres instance
(1)  we will update our code to use that secret .

api/v1alpha1/search_types.go
jlpadilla: Creating the configmap is easy, but it becomes another step and another object to track, backup, etc.

jlpadilla (approved):
/lgtm

Approving with a couple concerns with the API, but this is great progress!

README.md
```diff
@@ -41,17 +71,15 @@ If you want to replace any PR images for any of the search components , you can
...
 
-  ### 10. Verify for following search pods are running
+  ###  9. Verify for following search pods are running
               search-api
               search-collector
               search-indexer
               search-postgres
                        
-  # Installing using Bundle Image                          
+
```
dockerymick, line 84: Similar to the other comments. remove the number out of the section title

dockerymick (commented):
@xrajesh @jlpadilla

PROJECT
```diff
@@ -12,8 +12,8 @@ resources:
     namespaced: true
   controller: true
   domain: open-cluster-management.io
-  group: cache
-  kind: OCMSearch
-  path: github.com/stolostron/search-v2-operator/api/v1
-  version: v1
+  group: search
+  kind: Search
+  path: github.com/stolostron/search-v2-operator/api/v1alpha1
```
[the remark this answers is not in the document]
jlpadilla: This is dev preview and not included in the product yet, but let's discuss how we can start creating this content to be merged when we GA in 2.6.

PR #13, opened 2022-03-15
Watch resources and customize

xrajesh:


controllers/common.go
```diff
@@ -13,12 +34,404 @@ func getServiceAccountName() string {
 	return "search-serviceaccount"
 }
 
-func getImagePullSecret() string {
+func getImagePullSecretName() string {
 	return "search-pull-secret"
 }
+
```
jlpadilla, line 40: We could have a common function to get the deployment from name like this:
```go
func getDeploymentConfig(name string, instance *searchv1alpha1.Search) DeploymentConfig {
    switch name {
    case apiDeploymentName:
        return instance.Spec.Deployments.API
    case collectorDeploymentName:
        return instance.Spec.Deployments.Collector
    case indexerDeploymentName:
        return instance.Spec.Deployments.Indexer
    case postgresDeploymentName:
        return instance.Spec.Deployments.Database
    }
    return nil
} 
```

Then the functions below can be simplified like this:
```go
func getNodeSelector(deploymentName string, instance *searchv1alpha1.Search) map[string]string {
        deploymentConfig := getDeploymentConfig(deploymentName, instance)
        if deploymentConfig.NodeSelector != nil {
            return deploymentConfig.NodeSelector
        }
	var result map[string]string
        return result
}
```

controllers/common.go
```diff
@@ -13,12 +34,404 @@ func getServiceAccountName() string {
...
+func getReplicaCount(deploymentName string, instance *searchv1alpha1.Search) *int32 {
+
+	count := int32(1)
+	if c, ok := defaultReplicaMap[deploymentName]; ok {
+		count = c
+	}
+	switch deploymentName {
+	case apiDeploymentName:
+
+		if instance.Spec.Deployments.API.ReplicaCount != 0 {
```
jlpadilla, line 270: Should the API enforce `ReplicaCount > 0` ?

controllers/common.go
```diff
@@ -13,12 +34,404 @@ func getServiceAccountName() string {
...
+		corev1.ResourceMemory: memory,
+	}
+}
+
+func getReplicaCount(deploymentName string, instance *searchv1alpha1.Search) *int32 {
+
+	count := int32(1)
+	if c, ok := defaultReplicaMap[deploymentName]; ok {
+		count = c
+	}
```
jlpadilla, line 266: Does this account for `AvailabilityMode (High or Basic)`?

controllers/common.go
```diff
@@ -13,12 +34,404 @@ func getServiceAccountName() string {
...
+		if instance.Spec.Deployments.Database.Resources != nil {
+			return true
+		}
+	}
+	return false
+}
+
+func (r *SearchReconciler) createOrUpdateConfigMap(ctx context.Context, cm *corev1.ConfigMap) (*reconcile.Result, error) {
+	found := &corev1.ConfigMap{}
+	err := r.Get(context.TODO(), types.NamespacedName{
```
jlpadilla, line 369: Use the context that gets passed to the function `ctx`.  There's several cases below.

controllers/common.go
```diff
@@ -13,12 +34,404 @@ func getServiceAccountName() string {
...
+		}
+	case postgresDeploymentName:
+		if instance.Spec.Deployments.Database.Resources != nil {
+			return true
+		}
+	}
+	return false
+}
+
+func (r *SearchReconciler) createOrUpdateConfigMap(ctx context.Context, cm *corev1.ConfigMap) (*reconcile.Result, error) {
```
jlpadilla, line 367: Only create case is covered. Bug or misleading function name?

controllers/common.go
```diff
@@ -13,12 +34,404 @@ func getServiceAccountName() string {
...
+		if err != nil {
+			log.Error(err, "Could not create deployment")
+			return &reconcile.Result{}, err
+		}
+	}
+	if err := r.Update(context.TODO(), deploy); err != nil {
+		log.Error(err, "Could not update deployment")
+		return &reconcile.Result{}, err
+	}
+	log.V(2).Info("Created %s deployment", deploy.Name)
```
jlpadilla, line 401: This log could be inaccurate when the deployment is updated.

controllers/common.go
```diff
@@ -13,12 +34,404 @@ func getServiceAccountName() string {
...
+	return nil, nil
+}
+
+func (r *SearchReconciler) createOrUpdateService(ctx context.Context, svc *corev1.Service) (*reconcile.Result, error) {
```
jlpadilla, line 405: Missing update path.  Same with next function.

controllers/defaults.go
```diff
@@ -0,0 +1,64 @@
...
+	default_Collector_MemoryRequest = "64Mi"
+
+	default_Postgres_CPURequest    = "25m"
+	default_Postgres_MemoryLimit   = "4Gi"
+	default_Postgres_MemoryRequest = "128Mi"
+
+	default_API_Replicas       = 2
+	default_Indexer_Replicas   = 1
+	default_Collector_Replicas = 1
+	default_Postgres_Replicas  = 1
```
jlpadilla, line 24: - Why API defaults to 2?
- Can Postgres have more than 1 replica with our current config?

controllers/common.go
jlpadilla: Would it help to also log the entire object **when verbosity is high?**
```go
// We could add both logs at different verbosity levels.
log.V(2).Info("Created %s deployment", deploy.Name)
log.V(9).Info("Created deployment. %+v, deploy)
```

controllers/defaults.go
xrajesh: We have 2 pods in search V1 - Going with that.
We cannot have more than 1 replica for search postgres in current config

controllers/common.go
xrajesh: AvailabilityMode - is not accounted in this PR .

controllers/defaults.go
jlpadilla: - In V1, the API deployment is the only one that honors the AvailabilityMode set in MCH, so that's why it defaults to 2.
- Indexer should do the same as API.
- Maybe we need to add the concept of `max_replicas` ?  Postgres and collector MUST only have 1 replica, so it will cause problems if user tries to override.  I believe this PR is ignoring replicas for those pods, that's ok, but need to make sure we addd a status message or it will be confusing.

xrajesh: I will do default , to 1 for all 4 pods . We will have a validation for the ReplicaCount to be > than 0 . 
max_replicas - concept, needs some discussion.

controllers/common.go
xrajesh: Added Update for Service , Renamed the function name to createConfigMap

controllers/common.go
xrajesh: Fixed it

controllers/common.go
xrajesh: Misleading name - fixed it .

controllers/common.go
xrajesh: Fixed it.

controllers/common.go
xrajesh: Made it >=0

controllers/common.go
xrajesh: Fixed it

controllers/common.go
```diff
@@ -13,12 +34,316 @@ func getServiceAccountName() string {
...
+func hasDeployments(instance *searchv1alpha1.Search) bool {
+	return instance.Spec.Deployments.DeepCopy() != nil
+
+}
+
+func isDeploymentCustomized(deploymentName string, instance *searchv1alpha1.Search) bool {
+	if !hasDeployments(instance) {
+		return false
+	}
+	switch deploymentName {
```
jlpadilla, line 220: could use `getDeploymentConfig()` here.

api/v1alpha1/search_types.go
```diff
@@ -85,13 +85,15 @@ type SearchDeployments struct {
 
 type DeploymentConfig struct {
 	// +optional
+	// +kubebuilder:validation:Minimum:=0
```
jlpadilla, line 88: should this be `Minimum:=1` instead of 0?

jlpadilla (approved):
/lgtm

Added a few minor comments.

api/v1alpha1/search_types.go
xrajesh: Yes - this could go 1 . 0 does not add any value.

controllers/common.go
xrajesh: Yes updated this.

controllers/common.go
```diff
@@ -13,12 +34,316 @@ func getServiceAccountName() string {
...
+	deploymentConfig := getDeploymentConfig(deploymentName, instance)
+	if deploymentConfig.ReplicaCount != 0 {
+		return &deploymentConfig.ReplicaCount
+	}
+	return &count
+
+}
+func getImageSha(deploymentName string, instance *searchv1alpha1.Search) string {
+	switch deploymentName {
+	case apiDeploymentName:
```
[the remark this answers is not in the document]
xrajesh: We need a map matching names - will bring in next PR .

PR #18, opened 2022-03-28
check for deployment updates

xrajesh:
Signed-off-by: xavier dharmaiyan <xavier@redhat.com>

### DIFF
--- a/controllers/common.go
+++ b/controllers/common.go
@@ -8,6 +8,7 @@ import (
 	searchv1alpha1 "github.com/stolostron/search-v2-operator/api/v1alpha1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


controllers/common.go
```diff
@@ -312,14 +321,24 @@ func (r *SearchReconciler) createSecret(ctx context.Context, secret *corev1.Secr
...
-			return &reconcile.Result{}, err
+	if err != nil {
+		if errors.IsNotFound(err) {
+			err = r.Create(ctx, secret)
+			if err != nil {
+				log.Error(err, "Could not create secret")
+				return &reconcile.Result{}, err
+			}
+			log.V(2).Info("Created %s secret", secret.Name)
+			log.V(9).Info("Created secret %+v", secret)
```
jlpadilla, line 332: This will log the contents of the secret which could be a security problem!

xrajesh: Even with V(9) ? @jlpadilla

PR #20, opened 2022-03-30
Initial RBAC documentation

jlpadilla:
Signed-off-by: Jorge Padilla <jpadilla@redhat.com>

Initial documentation for Search V2 RBAC implementation.

### Here's how to access the pretty view of this document.
image


docs/RBAC.md
```diff
@@ -0,0 +1,97 @@
+# Role Based Access Control (RBAC)
+
+### Status: DRAFT
+**This feature is not fully implemented. This document describes the desired imlpementation for Odessey (search v2).**
+
+The search service collects data using a service account with wide cluster access and stores all resources in the database. The API must enforce that results for each user (or service account) only contain resources that they are authorized to access.
+
+## Access to the API
```
bjoydeep, line 8: Should we say that this is `Query API` . As you know, we have another API to post data.

docs/RBAC.md
```diff
@@ -0,0 +1,97 @@
...
+> 4. Validation at the service? I want to avoid this if possible.
+
+## Enforcing RBAC on results
+
+The API authenticates the user (or service account) and impersonates the user to obtain their access rules.
+
+> **DISCUSSION:** What's the correct way to impersonate the user? Currently, we are just using their token, which isn't ideal. The request should made "by search on behalf of user".
+> 
+> I expect this to be similar to authorizing an app on Github to take actions on my behalf.
+
```
bjoydeep, line 27: There are 2 scenarios in the long run. They may not be applicable today. See if you concur:
1. user makes a synchronous call and needs an answer. The user token is available
2. Another service is making calls in the background may be on behalf of a user request that was initiated a while back. There is no token in this case (we may not have a case for this now - but it usually one of these comes up specially we come up with long running API - given the fact that we have to gather data from a large DB). We will have to trust the service in this case.
In other words, no matter what the implementation is, we will have to evaluate the rights of the user and perhaps also rights of other services who may call us w/o any user credentials. Does this make sense?

jlpadilla: Ok, let's talk scenario #2.  I believe that we can cover it using impersonation.  We'll need to authenticate and trust both the App and the user, then impersonate the user.

https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation

docs/RBAC.md
```diff
@@ -0,0 +1,152 @@
...
+Building the user's authorization rules requires a lot of API requests because these APIs are scoped to a single resource or namespace. We must cache the results to minimize the impact on the Kubernetes API server.
+
+Data is cached within the API pod (golang). The Kubernetes Service load balancer is configured with `SessionAffinity: ClusterIP` to route requests from a given user to the same API pod instance. This configuration eliminates the need for a shared cache across pods.
+
+The default time-to-live (TTL) is 10 minutes. Each incoming request from the user resets the cache expiration.
+
+We watch the Kubernetes resources with the RBAC definitions and invalidate the cache when any of these resources change. Invalidated caches will get rebuilt on the next incoming request for each user. This adds a delay to the request, but reduces the pressure on the Kube API server. 
+
+- **Role** 
+    - Invalidate cached rules for the affected namespace. **[TODO: Confitm that Role is always scoped to a single namespace.]**
```
Anxhela21, line 115: **[TODO: Confitm  ----> **[TODO: **Confirm**

jlpadilla:
Moved to PR #28 because of merge conflicts.

docs/RBAC.md
jlpadilla: updated, thanks

PR #21, opened 2022-03-31
Add search addons for managed clusters

xrajesh:


main.go
```diff
@@ -74,10 +76,15 @@ func main() {
...
 		LeaderElectionID:       "b648e39a.open-cluster-management.io",
 	})
+
 	if err != nil {
 		setupLog.Error(err, "unable to start manager")
 		os.Exit(1)
 	}
+	if err := addonv1alpha1.AddToScheme(mgr.GetScheme()); err != nil {
+		log.Error(err, "unable to add to scheme")
+		os.Exit(1)
```
jlpadilla, line 86: Can we avoid exiting on this error?  The operator will be significantly degraded, but we can still perform the install activities.  Is there any possible way to recover from this error with a retry?

jlpadilla (approved):
/lgtm
/hold

main.go
xrajesh: @jlpadilla Addressed this in the latest change - could you please review it ?

PR #26, opened 2022-04-25
Install addon after search CR is created

xrajesh:


controllers/common.go
```diff
@@ -54,6 +56,17 @@ func newMetadataEnvVar(name, key string) corev1.EnvVar {
 	}
 }
 
+func IsPaused(annotations map[string]string) bool {
+	if annotations == nil {
```
SherinV, line 60: Better to use len(annotations) to check if it is empty as it is a map.

controllers/create_clustermanagementaddon.go
```diff
@@ -50,4 +51,9 @@ func newClusterManagementAddOn(instance *searchv1alpha1.Search) *addonv1alpha1.C
 			},
 		},
 	}
+	err := controllerutil.SetControllerReference(instance, cma, r.Scheme)
+	if err != nil {
+		log.V(2).Info("Could not set control for search-collector ClusterManagementAddOn", err)
```
SherinV, line 56: Do we need to return error and retry here?

addon/addon.go
```diff
@@ -166,3 +167,29 @@ func NewAddonManager(kubeConfig *rest.Config) (addonmanager.AddonManager, error)
...
+	} else {
+		klog.Info("starting search addon manager")
+		err = addonMgr.Start(ctx)
+		if err != nil {
+			klog.Error(err, "unable to start a new  addon manager", "controller", "SearchOperator")
+		}
+	}
+}
+
+func CreateAddonOnce(ctx context.Context, instance *searchv1alpha1.Search) {
```
jlpadilla, line 189: Can you add a quick description for this function?  I'm looking for why "Once" ?

controllers/common.go
```diff
@@ -21,6 +22,7 @@ const (
 	collectorDeploymentName = "search-collector"
 	indexerDeploymentName   = "search-indexer"
 	postgresDeploymentName  = "search-postgres"
+	AnnotationSearchPause   = "search-pause"
```
jlpadilla, line 25: How about "operator-pause" to keep it generic?
How does mch calls it? Is it mch-pause?

controllers/common.go
```diff
@@ -54,6 +56,17 @@ func newMetadataEnvVar(name, key string) corev1.EnvVar {
 	}
 }
 
+func IsPaused(annotations map[string]string) bool {
+	if annotations == nil {
+		return false
+	}
+	if annotations[AnnotationSearchPause] != "" &&
```
jlpadilla, line 63: Do you need to check for `annotations[AnnotationSearchPause] != nil` ?

controllers/create_clustermanagementaddon.go
jlpadilla, line 56: Is `V(2)` correct here? Message sounds like a "possible problem".

controllers/search_controller.go
```diff
@@ -62,6 +65,12 @@ func (r *SearchReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		}
 		return reconcile.Result{}, err
 	}
+
+	// Do not reconcile objects if this instance of search is labeled "paused"
+	if IsPaused(instance.GetAnnotations()) {
+		log.Info("Search reconciliation is paused. Nothing more to do.")
```
jlpadilla, line 71: - Avoid the word "Search" in case we ever need to rename this.
- Should we add a hint to help resume the operator?
```suggestion
		log.Info("Reconciliation is paused because the annotation 'operator-pause: true' was found.")
```

addon/addon.go
xrajesh: Done

controllers/common.go
xrajesh: yes - `mch-pause` .  Keeping same naming convention.

controllers/common.go
xrajesh: No , we will get a false returned in that case.

controllers/create_clustermanagementaddon.go
xrajesh: No a serious error , we dont want to retry here .

xrajesh: Made as a error log

controllers/search_controller.go
xrajesh: Done.

PR #28, opened 2022-04-27
RBAC documentation

jlpadilla:
Signed-off-by: Jorge Padilla <jpadilla@redhat.com>

Initial documentation for Search V2 RBAC implementation.

### Here's how to access the pretty view of this document.
image

### DIFF
--- a/docs/RBAC.md
+++ b/docs/RBAC.md
@@ -1,0 +1,152 @@
+# Role Based Access Control (RBAC)
+
+### Status: DRAFT (April 22, 2022)
+**This feature is not fully implemented. This document describes the desired imlpementation for Odyssey (search v2).**
+
+The search service collects data using a service account with wide cluster access and stores all resources in the database. The Search API must enforce that results for each user (or service account) only contain resources that they are authorized to access.
+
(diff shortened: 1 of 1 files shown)


jlpadilla:
I want to merge this by tomorrow 4/28. We can continue improving it after. Please add your review @bjoydeep @xrajesh @Anxhela21 @SherinV 

(I had to create a new PR to resolve conflicts)

Anxhela21 (approved):
/lgtm
/hold

jlpadilla:
@bjoydeep @xrajesh @SherinV Last chance to review before I merge this.

docs/RBAC.md
```diff
@@ -0,0 +1,152 @@
...
+> Options:
+> 1. Kube API server extension - This can add additional load to the kube API server.
+> 2. Management Ingress - This is being deprecated.
+> 3. Validate access at the service - Provides the most flexibility, but how will it leverage kubernetes implementation?
+
+## Authenticate the user 
+
+The Search API authenticates the user (or service account) and impersonates the user to obtain their access rules.
+We cache the token validation for a short period (60 seconds) to reduce requests to the kube api. Tokens have a short expiration, so we must revalidate often. **Note that** this time-to-live period is independent of the cached rules, which is longer at 10 minutes of inactivity.
+
```
xrajesh, line 32: Is this validation time 60s - this is configurable ?

docs/RBAC.md
```diff
@@ -0,0 +1,152 @@
...
+1. [Rules for resources in the hub cluster](#hub-cluster)
+2. [Rules for resources in managed clusters](#managed-clusters)
+
+> Use [User Impersonation](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation) when sending API requests on behalf of the end user.
+```
+CHANGE IN V2. [Does not affect API results.]
+V1 passes the user's token directly to the API, but impersonation is desired.
+This change doesn't change the semantic of the API.
+```
+
```
xrajesh, line 46: What is the user which runs the query - with impersonation ?

docs/RBAC.md
```diff
@@ -0,0 +1,152 @@
...
+### 1. Hub cluster
+
+The search results must include **exactly** the same resources users are authorized to list using kubectl, oc, or the kubernetes API on the OpenShift cluster hosting the ACM Hub.
+
+We request all the authorization rules for the user and [cache](#cache) the results.
+> 1. Get all resources in the cluster that support `list` and `watch`. This is shared across all users.
```
xrajesh, line 52: Why do we need watch ? list isnt enough ?

docs/RBAC.md
```diff
@@ -0,0 +1,152 @@
...
+```
+
+Find the managed clusters that the user is authorized to view and [cache](#cache) the results.
+> 1. Get all the namespaces associated with a managed cluster. We do this once for all users.
+>       - CLI: `oc get ManagedClusters`
+> 2. Get all projects (namespaces) for the user. (We already have this data.)
+>       - CLI: `oc projects --as=<user>`
+>       - API: [ProjectList](https://docs.okd.io/3.9/rest_api/apis-project.openshift.io/v1.Project.html#Get-apis-project.openshift.io-v1-projects)
+> 3. Build a list of all the managed clusters visible to the user.
+>       - UNION of results from steps 1 and 2.
```
xrajesh, line 98: UNION or INTERSECTION ?

docs/RBAC.md
jlpadilla: Updated to `(default: 60 seconds)` to be a little clearer.

docs/RBAC.md
jlpadilla: Updated per Slack discussion.

docs/RBAC.md
jlpadilla: The collector only "collects" resources that support `watch` so if the resource doesn't support watch we ignore it because we know it's not in our data.
Similarly, if the resource doesn't support `list`, we know that users won't have permission to list because the operation isn't supported by the resource.

docs/RBAC.md
jlpadilla: Thanks, I meant INTERSECTION.  Updated.

xrajesh (approved):

SherinV (approved):

PR #35, opened 2022-05-09
Create pvc if user wants postgres persistence 

xrajesh:


api/v1alpha1/search_types.go
```diff
@@ -15,6 +15,7 @@ const (
 	// HABasic stands up most app subscriptions with a replicaCount of 1
 	HABasic AvailabilityType = "Basic"
 	// HAHigh stands up most app subscriptions with a replicaCount of 2
+	// Not supported for Dev preview
```
jlpadilla, line 18:
```suggestion
	// Not supported for development preview.
```

api/v1alpha1/search_types.go
```diff
@@ -52,13 +53,22 @@ type SearchSpec struct {
 	AvailabilityConfig AvailabilityType `json:"availabilityConfig,omitempty"`
 
 	// +optional
-	// Control list of Kubernetes resources indexed by search-collector
-	AllowDenyResources FilterSpec `json:"allowDenyResources,omitempty"`
-
-	// +optional
+	// Not part of dev preview
 	// Kubernetes secret name containing user provided db secret
 	// Secret should contain connection parameters [db_host, db_port, db_user, db_password, db_name, ca_cert]
```
jlpadilla, line 58: Match style of `HAHigh`.
```suggestion
	// Kubernetes secret name containing user provided db secret
	// Secret should contain connection parameters [db_host, db_port, db_user, db_password, db_name, ca_cert]
	// Not supported for development preview.
```

api/v1alpha1/search_types.go
```diff
@@ -52,13 +53,22 @@ type SearchSpec struct {
...
 	ExternalDBInstance string `json:"externalDBInstance,omitempty"`
+
+	// +optional
+	//ImagePullSecret
```
jlpadilla, line 62: **Super minor format:** several comments are missing a space after `//`.

I wonder if there's a lint rule for this.

api/v1alpha1/search_types.go
```diff
@@ -94,69 +100,33 @@ type DeploymentConfig struct {
...
-
-	// NodeSelector to schedule on nodes with matching labels
-	// +optional
-	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
-
-	//Proxy config , if remote collectors need to override
-	// +optional
-	ProxyConfig map[string]string `json:"proxyConfig,omitempty"`
+	// Container Arguments
+	Arguments []string `json:"arguments,omitempty"`
```
jlpadilla, line 108: Should we also add environment variables to be passed to the container?

api/v1alpha1/search_types.go
```diff
@@ -94,69 +100,33 @@ type DeploymentConfig struct {
...
 }
 
 type StorageSpec struct {
-	// +optional
 	// name of the storage class
 	StorageClassName string `json:"storageClassName,omitempty"`
 	// +optional
-	// storage capacity
+	// +kubebuilder:default:="10Gi"
```
jlpadilla, line 115: I don't know how to do it from here, but in V1 we define a regular expression to validate the values. Should we add that here?

controllers/common.go
```diff
@@ -83,28 +83,70 @@ func IsPaused(annotations map[string]string) bool {
...
+	case indexerDeploymentName:
+		if instance.Spec.Deployments.Indexer.Arguments != nil {
+			return instance.Spec.Deployments.Indexer.Arguments
+		}
+	case postgresDeploymentName:
+		if instance.Spec.Deployments.Database.Arguments != nil {
+			return instance.Spec.Deployments.Database.Arguments
+		}
+	}
+	log.V(2).Info("Unknown deployment %s ", deploymentName)
```
jlpadilla, line 142: We want to log this only when none of the case conditions match.
```suggestion
	default:
		log.V(2).Info("Unknown deployment %s ", deploymentName)
	}
```

controllers/create_pvc.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+func (r *SearchReconciler) configurePVC(ctx context.Context, instance *searchv1alpha1.Search) bool {
+	if err := r.createPVC(ctx, instance); err != nil {
+		return false
+	}
+	return true
+}
+
+func (r *SearchReconciler) createPVC(ctx context.Context, instance *searchv1alpha1.Search) error {
```
jlpadilla, line 22: What would happen if the PVC exists with a different size? Should this logic be updated to create or update PVC?

api/v1alpha1/search_types.go
xrajesh: Size type is  *resource.Quantity - That will auto check , you cant include invalid values

controllers/common.go
xrajesh: Right - but the valid cases would have returned . This  line will only execute for the ones we dont know about.

controllers/create_pvc.go
xrajesh: Per discussion with @jlpadilla @bjoydeep we are not planning to support PVC resize.

jlpadilla: I'm OK if we don't support PVC resize, but how do we prevent users from changing the size value after the CR is created?

api/v1alpha1/search_types.go
jlpadilla: Good, auto check is much better! :-)

controllers/create_pvc.go
xrajesh: Users can change it , but it will not have any impact . If Users really want to change size, they have to bring down search and Update the PVC size and restart search .

controllers/common.go
xrajesh: The operators are using this package to log "sigs.k8s.io/controller-runtime/pkg/log" - That does not have the Infof . I see folks using Info . @jlpadilla

controllers/common_test.go
```diff
@@ -355,5 +365,47 @@ func TestPostgresCustomization(t *testing.T) {
 	if actual_image_sha != "quay.io/test-image:007" {
 		t.Errorf("ImageOverride with incorrect image")
 	}
+	actual_args := getContainerArgs(testFor, instance)
+	if actual_args == nil || len(actual_args) != 1 || actual_args[0] != "arg1" {
+		t.Errorf("Incorrect Args parsed")
```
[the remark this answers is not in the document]
xrajesh: Done @jlpadilla

controllers/create_pvc.go
jlpadilla: ok, I understand the challenge and I don't want to block this PR, so let's merge but keep this discussion open.

For the user it will be frustrating that their changes don't take effect. I'd like to see an error state or something that indicates that we are unable to apply the requested desired state.

jlpadilla (approved):
/lgtm
/hold

PR #40, opened 2022-06-06
lower case kind in properties

SherinV:
Signed-off-by: Sherin Varughese <shvarugh@redhat.com>

**Related Issue:**  stolostron/backlog#22013

### Description of changes
Convert kind to lowercase


SherinV:
case-insensitive comparison for kind handled in https://github.com/stolostron/search-v2-api/pull/52

PR #59, opened 2022-07-26
Update to RBAC Doc - managed cluster cache

Anxhela21:
Signed-off-by: Anxhela <acoba@redhat.com>

### Description of changes
Update to the RBAC doc to better reflect true method used to get namespaces. Similar to the hub, "users don't always have visibility to all the projects where they can list resources. Unfortunately we'll need to query all namespaces to get accurate results in the next step"


jlpadilla (approved):
/lgtm

PR #63, opened 2022-08-04
Remove clustermanagementaddon CR search-collector if search is removed

xrajesh:

### DIFF
--- a/controllers/cleanup.go
+++ b/controllers/cleanup.go
@@ -1,0 +1,34 @@
+// Copyright Contributors to the Open Cluster Management project
+package controllers
+
+import (
+	"context"
+	"time"
+
+	searchv1alpha1 "github.com/stolostron/search-v2-operator/api/v1alpha1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	addonapiv1alpha1 "open-cluster-management.io/api/addon/v1alpha1"
+)
+
+func (r *SearchReconciler) deleteClusterManagementAddon(instance *searchv1alpha1.Search) error {
+	log.V(2).Info("Deleting ClusterManagementAddon search-collector")
+	cma := &addonapiv1alpha1.ClusterManagementAddOn{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "ClusterManagementAddon",
+			APIVersion: "addon.open-cluster-management.io",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "search-collector",
+			Namespace: instance.GetNamespace(),
+		},
+	}
+	err := r.Delete(context.TODO(), cma)
+	if err != nil && !errors.IsNotFound(err) {
+		log.Error(err, "Failed to delete ClusterManagementAddon", "name", cma)
+		return err
+	}
+	time.Sleep(1 * time.Second)
+	log.V(2).Info("ClusterManagementAddon search-collector deleted", "name", cma)
+	return nil
+}
--- a/controllers/cleanup.go
+++ b/controllers/cleanup.go
@@ -3,7 +3,6 @@ package controllers
 
 import (
 	"context"
-	"time"
 
 	searchv1alpha1 "github.com/stolostron/search-v2-operator/api/v1alpha1"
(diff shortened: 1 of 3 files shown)


zhiweiyin318:
I have some confusion why delete `clusterManagementAddon` when search CR is deleted?
`clusterManagementAddon` on the hub cluster is to say search-collector addon operator (search-v2-operator) has been installed on the hub, you can enable the search-collector addon agent on your managed cluster by search-collector managedClusterAddon CR.
I think `clusterManagementAddon` should be deleted only when the search-collector addon operator (search-v2-operator) is removed.

xrajesh:
@zhiweiyin318 Thank you for your feedback. We are creating `clusterManagementAddon` instance named `search-collector` in the Reconcile function , which executes only after a `search` instance named `search-v2-operator` is created. So once we remove the search instance, we can also remove  `clusterManagementAddon` instance named `search-collector` right , because - next time a search CR instance is created we create the  `clusterManagementAddon`  again. Currently we are creating the search-v2-operator from a bundle, and the search-v2-operator does not do anything until we have a search CR @jlpadilla FYI.

controllers/search_controller.go
```diff
@@ -180,6 +191,48 @@ func (r *SearchReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			// that we can retry during the next reconciliation.
+			if err := r.finalizeSearch(instance); err != nil {
+				return err
+			}
+			// Remove searchFinalizer. Once all finalizers have been
+			// removed, the object will be deleted.
+			controllerutil.RemoveFinalizer(instance, searchFinalizer)
+			r.Update(ctx, instance)
+			log.Info("Finalizer removed from search CR")
+		}
```
zhiweiyin318, line 226: should return here ?

xrajesh: Done .

controllers/search_controller.go
```diff
@@ -62,17 +64,30 @@ func (r *SearchReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
 	if IsPaused(instance.GetAnnotations()) {
 		log.Info("Reconciliation is paused because the annotation 'search-pause: true' was found.")
 		return ctrl.Result{}, nil
 	}
+
+	// Setup finalizers
+	deleted, err := r.setFinalizer(ctx, instance)
+	if err != nil || deleted {
+		if deleted {
+			log.V(2).Info("Search Instance deleted, reque request", err)
```
SherinV, line 82: requeue*

SherinV: Same on line 84

controllers/search_controller.go
```diff
@@ -180,6 +195,50 @@ func (r *SearchReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			// Run finalization logic for searchFinalizer. If the
+			// finalization logic fails, don't remove the finalizer so
+			// that we can retry during the next reconciliation.
+			if err := r.finalizeSearch(instance); err != nil {
+				return false, err
+			}
+			// Remove searchFinalizer. Once all finalizers have been
+			// removed, the object will be deleted.
+			controllerutil.RemoveFinalizer(instance, searchFinalizer)
+			r.Update(ctx, instance)
```
SherinV, line 228: log and return error if update fails?

controllers/cleanup.go
```diff
@@ -0,0 +1,34 @@
...
+			Name:      "search-collector",
+			Namespace: instance.GetNamespace(),
+		},
+	}
+	err := r.Delete(context.TODO(), cma)
+	if err != nil && !errors.IsNotFound(err) {
+		log.Error(err, "Failed to delete ClusterManagementAddon", "name", cma)
+		return err
+	}
+	time.Sleep(1 * time.Second)
```
jlpadilla, line 31: Why do we need to sleep?

jlpadilla: If it's for the retry logic, maybe only sleep if there's and error?

xrajesh: Had that only as a safety measure - Did not want to start a reconcile on a instance which is SET to delete .I am removing it for now , if we see any issues , we can check again.

SherinV (approved):

jlpadilla (approved):
/lgtm
/hold

PR #76, opened 2022-09-28
perf improvements and klusterlet customizations

xrajesh:
Signed-off-by: Xavier Dharmaiyan <xavier@redhat.com>

**Related Issue:**  stolostron/backlog#18606 and stolostron/backlog#25847

### Description of changes
- Adds annotations to support override of default configuration in add-on deployment.
- Reduce scope of resource watches to reduce memory usage.

### DIFF
--- a/addon/manifests/chart/templates/deployment.yaml
+++ b/addon/manifests/chart/templates/deployment.yaml
@@ -24,7 +24,7 @@ spec:
       - name: collector
         image: "{{ .Values.global.imageOverrides.search_collector }}"
         imagePullPolicy: {{ .Values.global.imagePullPolicy }}
-                {{- if .Values.userargs.containerArgs }}
+        {{- if .Values.userargs.containerArgs }}
         args:
           - '{{ .Values.userargs.containerArgs }}'
         {{- end }}
--- a/addon/manifests/chart/templates/deployment.yaml
+++ b/addon/manifests/chart/templates/deployment.yaml
@@ -24,6 +24,10 @@ spec:
       - name: collector
         image: "{{ .Values.global.imageOverrides.search_collector }}"
         imagePullPolicy: {{ .Values.global.imagePullPolicy }}
+                {{- if .Values.userargs.containerArgs }}
+        args:
+          - '{{ .Values.userargs.containerArgs }}'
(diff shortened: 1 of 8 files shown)


addon/manifests/chart/templates/deployment.yaml
```diff
@@ -24,6 +24,10 @@ spec:
       - name: collector
         image: "{{ .Values.global.imageOverrides.search_collector }}"
         imagePullPolicy: {{ .Values.global.imagePullPolicy }}
+                {{- if .Values.userargs.containerArgs }}
```
jlpadilla, line 27: indentation might be wrong.

xrajesh: Fixed it

controllers/search_controller.go
```diff
@@ -205,12 +203,14 @@ func (r *SearchReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-		Watches(&source.Kind{Type: &rbacv1.ClusterRoleBinding{}}, &handler.EnqueueRequestForObject{}, builder.WithPredicates(pred)).
-		Watches(&source.Kind{Type: &rbacv1.ClusterRole{}}, &handler.EnqueueRequestForObject{}, builder.WithPredicates(pred)).
+		Watches(&source.Kind{Type: &appsv1.Deployment{}}, &handler.EnqueueRequestForOwner{
+			IsController: true,
+			OwnerType:    &searchv1alpha1.Search{},
+		}, builder.WithPredicates(pred)).
+		Watches(&source.Kind{Type: &corev1.Secret{}}, &handler.EnqueueRequestForOwner{
+			IsController: true,
+			OwnerType:    &searchv1alpha1.Search{},
+		}, builder.WithPredicates(pred)).
```
jlpadilla, line 213: Curious why we aren't watching other resources anymore.
Would the result be that we could get to a broken state if any of the resources not watched change?

jlpadilla (approved):
/lgtm
/hold

controllers/search_controller.go
xrajesh: Reducing the reconciliation only to Deployment changes and secrets .

PR #88, opened 2022-11-11
Add status to Search CR

SherinV:
**Related Issue:**  https://issues.redhat.com/browse/ACM-2005

### Description of changes
Add status update for Search CR

### DIFF
--- a/controllers/common.go
+++ b/controllers/common.go
@@ -480,7 +480,8 @@ func updateStatusCondition(instance *searchv1alpha1.Search, podList *corev1.PodL
 	}
 	var podPresent bool // bool to check if status for this pod already exists in Search instance
 	for i, instanceCondition := range instance.Status.Conditions {
-		if instanceCondition.Type == readyType {
+		// replace only for "Ready" condition and if the podCondition is not empty
+		if instanceCondition.Type == readyType && (metav1.Condition{}) != podCondition {
 
 			podPresent = true // status for this pod already exists in Search instance
 			// replace instance with the latest condition for this pod
@@ -488,10 +489,12 @@ func updateStatusCondition(instance *searchv1alpha1.Search, podList *corev1.PodL
 			break
 		}
 	}
-	if !podPresent { // status for this pod doesn't exist in Search instance
+	// append if the podCondition is not empty
+	if !podPresent && (metav1.Condition{}) != podCondition { // status for this pod doesn't exist in Search instance
 		instance.Status.Conditions = append(instance.Status.Conditions,
 			podCondition)
 	}
+
 	return instance
 }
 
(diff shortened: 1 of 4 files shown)


controllers/search_controller.go
```diff
@@ -193,6 +202,17 @@ func (r *SearchReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *SearchReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	searchPods := predicate.Funcs{
```
xrajesh, line 205: @SherinV We can remove this ? Because we reconcile for all events ?

controllers/search_test.go
```diff
@@ -322,3 +320,316 @@ func TestSearch_controller(t *testing.T) {
...
+	indexerPod := buildPod("search-indexer-abc", runningCondition)
+	postGresPod := buildPod("search-postgres-abc", runningCondition)
+
+	s := scheme.Scheme
+	err := searchv1alpha1.SchemeBuilder.AddToScheme(s)
+	if err != nil {
+		t.Errorf("error adding search scheme: (%v)", err)
+	}
+
+	err = addonv1alpha1.AddToScheme(s)
```
xrajesh, line 365: Do we need this ?

controllers/search_controller.go
SherinV: Done

controllers/search_test.go
SherinV: Removed.

controllers/common.go
```diff
@@ -439,3 +439,67 @@ func (r *SearchReconciler) createSecret(ctx context.Context, secret *corev1.Secr
 func DeploymentEquals(current, new *appsv1.Deployment) bool {
 	return equality.Semantic.DeepEqual(current.Spec, new.Spec)
 }
+
+// update status condition in search instance
+func updateStatusCondition(instance *searchv1alpha1.Search, podList *corev1.PodList) *searchv1alpha1.Search {
+	var podCondition metav1.Condition
+	var readyType string
+
+	for _, searchPod := range podList.Items {
```
jlpadilla, line 448: Will this logic work when we add multiple replicas of these pods? Should we use the deployment's `MinimumReplicasAvailable` condition instead?

xrajesh: It will work for Multiple replicas of ,say search-api pods. I will check all pods status and if any one of them is not ready the status will have the failure information. We thought about using deployments replica count , but then we had to again find the pod and get the cause of the failure.

jlpadilla: thanks!

jlpadilla (approved):
/lgtm
/hold

PR #111, opened 2023-02-15
[ACM-1854] add prometheus config for scraping

SherinV:
**Related Issue:**  [ACM-1854](https://issues.redhat.com/browse/ACM-1854)

### Description of changes
- Add prometheus config for scraping metrics from search pods.


controllers/create_metrics_monitor.go
```diff
@@ -0,0 +1,159 @@
...
+		Namespace: smonitor.Namespace,
+	}, found)
+	if err != nil && errors.IsNotFound(err) {
+		err = r.Create(ctx, smonitor)
+		if err != nil {
+			log.Error(err, "Could not create servicemonitor "+smonitor.Name)
+			return &reconcile.Result{}, err
+		}
+		log.Info("Created servicemonitor" + smonitor.Name)
+		log.V(9).Info("Created  servicemonitor ", "name", smonitor.Name)
```
jlpadilla, line 156: These 2 logs print the same info.  Did we intend to print the entire `smonitor` object instead?
I noticed a similar pattern at the top of this file.

controllers/create_metrics_monitor.go
```diff
@@ -0,0 +1,159 @@
...
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "search-metrics-monitor",
+			Namespace: instance.GetNamespace(),
+		},
+		Rules: getRules(),
+	}
+	err := controllerutil.SetControllerReference(instance, cr, r.Scheme)
+	if err != nil {
+		log.Info("Could not set control for Role ", "name", "search-metrics-monitor")
```
jlpadilla, line 73: Should this be logged as a warning or error? Similar pattern in the functions below.

controllers/create_metrics_monitor.go
SherinV: Yes, thanks.

controllers/create_metrics_monitor.go
SherinV: I don't think so. For pre-existing functions, it is not treated as an error.

jlpadilla (approved):
/lgtm
/hold

PR #121, opened 2023-04-14
ACM-3108 add env var to search deployments

SherinV:
### Related Issue

https://issues.redhat.com/browse/ACM-3108

### Description of changes
- Pass env variables to deployments controlled by the search operator.

### DIFF
--- a/controllers/create_pgconfigmap.go
+++ b/controllers/create_pgconfigmap.go
@@ -20,7 +20,7 @@ func (r *SearchReconciler) PostgresConfigmap(instance *searchv1alpha1.Search) *c
 			Namespace: ns,
 		},
 	}
-	work_mem := r.GetDBConfig(context.TODO(), instance, "WORK_MEM")
+	work_mem := r.GetDBConfigFromSearchCR(context.TODO(), instance, "WORK_MEM")
 	data := map[string]string{}
 	data["postgresql.conf"] = `
 ssl = 'on'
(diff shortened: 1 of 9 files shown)


controllers/create_pgconfigmap.go
```diff
@@ -20,7 +20,7 @@ func (r *SearchReconciler) PostgresConfigmap(instance *searchv1alpha1.Search) *c
 			Namespace: ns,
 		},
 	}
-	work_mem := r.GetDBConfig(context.TODO(), instance, "WORK_MEM")
+	work_mem := r.GetDBConfigFromSearchCR(context.TODO(), instance, "WORK_MEM")
```
SherinV, line 23: We use this in line 46 to create work_memquery.
work_memquery := "psql -d search -U searchuser -c \"ALTER ROLE searchuser set work_mem='" + work_mem + "'\""
Do we need to do this for other variables that the user adds?

SherinV: Discussed in scrum and agreed that this will be a separate user story.

jlpadilla (approved):
/lgtm
/hold

Before merging, we need to confirm that the CRD updates were propagated to search-v2-operator-bundle and picked up by the installer.

SherinV:
> /lgtm /hold
> 
> Before merging, we need to confirm that the CRD updates were propagated to search-v2-operator-bundle and picked up by the installer.

Installer picked up changes in https://github.com/stolostron/multiclusterhub-operator/pull/1015

PR #144, opened 2023-09-27
ACM-7673 Delete legacy servicemonitors

SherinV:
### Related Issue

https://issues.redhat.com/browse/ACM-7673

### Description of changes
- Added cleanup to remove servicemonitors from openshift-monitoring namespace.

### DIFF
--- a/controllers/search_controller.go
+++ b/controllers/search_controller.go
@@ -242,6 +242,7 @@ func (r *SearchReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		cleanOnce.Do(cleanUpFunc)
 		//To retry if cleanup fails
 		if !cleanupComplete {
+			log.Error(err, "Failed to remove legacy ServiceMonitor setup - returning error ")
 			return ctrl.Result{}, err
 		}
 	}
--- a/controllers/search_controller.go
+++ b/controllers/search_controller.go
@@ -56,8 +56,6 @@ var log = logf.Log.WithName("searchoperator")
 var once sync.Once
 var cleanOnce sync.Once
 
-var cleanupComplete bool
-
 //+kubebuilder:rbac:groups=search.open-cluster-management.io,resources=searches,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=search.open-cluster-management.io,resources=searches/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=search.open-cluster-management.io,resources=searches/finalizers,verbs=update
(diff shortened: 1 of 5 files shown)


controllers/search_controller.go
```diff
@@ -230,6 +223,25 @@ func (r *SearchReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+	cleanUpFunc := func() {
+		// delete legacy servicemonitor setup
+		result, err = r.deleteLegacyServiceMonitorSetup(instance)
+		cleanupComplete = true
+		log.Info("cleanupComplete. Setting cleanupComplete to ", "cleanupComplete", cleanupComplete)
+		if result != nil {
+			cleanupComplete = false
+			log.Info("cleanup not complete. Setting cleanupComplete to ", "cleanupComplete", cleanupComplete)
+			log.Error(err, "Failed to remove legacy ServiceMonitor setup ")
+		}
```
jlpadilla, line 235: Move the success log to the `else{}` to avoid printing both messages, which can be confusing.

```suggestion
		if result != nil {
			cleanupComplete = false
			log.Info("cleanup not complete. Setting cleanupComplete to ", "cleanupComplete", cleanupComplete)
			log.Error(err, "Failed to remove legacy ServiceMonitor setup ")
		} else {
			log.Info("cleanupComplete. Setting cleanupComplete to ", "cleanupComplete", cleanupComplete)
		}
```

controllers/search_controller.go
```diff
@@ -230,6 +223,25 @@ func (r *SearchReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	once.Do(func() {
 		addon.CreateAddonOnce(ctx, instance)
 	})
+	cleanUpFunc := func() {
```
jlpadilla, line 226:
```suggestion
	// Starting with ACM 2.9, ServiceMonitors are created in the open-cluster-management namespace.
	// This migration function removes the service monitors that were previously created in the openshift-monitoring namespace.
	// We can remove this migration step after ACM 2.8 End of Life.
	cleanUpFunc := func() {
```

controllers/search_controller.go
```diff
@@ -230,6 +223,25 @@ func (r *SearchReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+		cleanupComplete = true
+		log.Info("cleanupComplete. Setting cleanupComplete to ", "cleanupComplete", cleanupComplete)
+		if result != nil {
+			cleanupComplete = false
+			log.Info("cleanup not complete. Setting cleanupComplete to ", "cleanupComplete", cleanupComplete)
+			log.Error(err, "Failed to remove legacy ServiceMonitor setup ")
+		}
+	}
+	if !cleanupComplete {
+		cleanOnce.Do(cleanUpFunc)
```
jlpadilla, line 238: `cleanupComplete` ensures that we cleanup once or retry if failed.
So then, is  `cleanOnce.Do(...)` redundant here?

jlpadilla: Does it retry on every reconcile?  I wonder if an unrecoverable error will continue to happen in a loop and create problems here.

SherinV: It retries every loop. Rewrote that part so that it doesn't retry. A pod restart will fix in case any artifact remains.

jlpadilla (approved):
/lgtm

PR #146, opened 2023-10-04
log as error only if error is not NotFound

SherinV:
### Related Issue

https://issues.redhat.com/browse/ACM-0000

### Description of changes
- Added ...

### DIFF
--- a/controllers/cleanup.go
+++ b/controllers/cleanup.go
@@ -88,7 +88,7 @@ func (r *SearchReconciler) deleteLegacyServiceMonitorSetup(instance *searchv1alp
 	if err = r.Delete(r.context, r.MetricsRoleBinding(instance)); err != nil && !errors.IsNotFound(err) {
 		log.Error(err, "Failed to remove RoleBinding")
 	}
-	if err == nil {
+	if err == nil || (err != nil && errors.IsNotFound(err)) {
 		log.Info("Deleted legacy ServiceMonitor Setup from openshift-monitoring namespace")
 	} else {
 		log.Info("Failed to remove legacy ServiceMonitor setup from openshift-monitoring namespace")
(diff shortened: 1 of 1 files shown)


controllers/cleanup.go
```diff
@@ -88,7 +88,7 @@ func (r *SearchReconciler) deleteLegacyServiceMonitorSetup(instance *searchv1alp
 	if err = r.Delete(r.context, r.MetricsRoleBinding(instance)); err != nil && !errors.IsNotFound(err) {
 		log.Error(err, "Failed to remove RoleBinding")
 	}
-	if err == nil {
+	if err == nil || (err != nil && errors.IsNotFound(err)) {
 		log.Info("Deleted legacy ServiceMonitor Setup from openshift-monitoring namespace")
```
SherinV, line 92: Should I remove this logging as error can happen in a previous step and might not reach here as others may succeed?

jlpadilla: Right, `err` will only have the result of the last delete operation. We can log the result of each resource because it will only happen once.

SherinV: Removed it.

controllers/cleanup.go
```diff
@@ -88,9 +88,4 @@ func (r *SearchReconciler) deleteLegacyServiceMonitorSetup(instance *searchv1alp
 	if err = r.Delete(r.context, r.MetricsRoleBinding(instance)); err != nil && !errors.IsNotFound(err) {
 		log.Error(err, "Failed to remove RoleBinding")
 	}
```
jlpadilla, line 90: Let's add a simple log that we could use to confirm that this function executed. We can infer that it was successful if we don't see any errors logged before.

```suggestion
	}
	log.Info("Done deleting legacy service monitors.");
```

SherinV: Done

jlpadilla (approved):
/lgtm

PR #153, opened 2023-12-19
ACM-9135: add install strategy

SherinV:
### Related Issue

https://issues.redhat.com/browse/ACM-9135

### Description of changes
Resolve issue:
```
#oc -n ocm logs  search-v2-operator-controller-manager-67c57b94dd-v7tld -c manager --tail 7
2023-12-18T21:37:36Z ERROR Reconciler error {"controller": "search", "controllerGroup": "search.open-cluster-management.io", "controllerKind": "Search", "Search":

{"name":"search-v2-operator","namespace":"ocm"}
, "namespace": "ocm", "name": "search-v2-operator", "reconcileID": "3a2b2e07-c72f-4318-aab4-85203faa2f0e", "error": "ClusterManagementAddOn.addon.open-cluster-management.io \"search-collector\" is invalid: spec.installStrategy.type: Unsupported value: \"\": supported values: \"Manual\", \"Placements\""}
```

PR #162, opened 2024-02-19
Removed installstrategy from CMA spec

dislbenn:
### Related Issue

https://issues.redhat.com/browse/ACM-9340

### Description of changes
- Removing install strategy from CMA spec. By default, the install strategy is type manual; therefore, we don't need to define it.

### DIFF
--- a/bundle/manifests/search-collector.clustermanagementaddon.yaml
+++ b/bundle/manifests/search-collector.clustermanagementaddon.yaml
@@ -9,5 +9,3 @@ spec:
       Collects cluster data to be indexed by search components on the hub
       cluster.
     displayName: Search Collector
-  installStrategy:
-    type: Manual


jlpadilla:
@dislbenn We recently added the `installStrategy` because it became a required field after updating the addon-framework go library. Here's the PR that added it. https://github.com/stolostron/search-v2-operator/pull/153

Maybe a better fix is to update the addon-framework api where this is causing problems. Otherwise this will just provide a temporary fix and the problem will reappear later.

FYI @SherinV

SherinV:
> @dislbenn We recently added the `installStrategy` because it became a required field after updating the addon-framework go library. Here's the PR that added it. #153
> 
> Maybe a better fix is to update the addon-framework api where this is causing problems. Otherwise this will just provide a temporary fix and the problem will reappear later.
> 
> FYI @SherinV

Yes, without the change Search failed to deploy. https://issues.redhat.com/browse/ACM-9135. Thanks.

dislbenn:
Thanks @jlpadilla @SherinV for the input. I will close this PR.

PR #174, opened 2024-04-23
Ignore gosec false positive

jlpadilla:
### Related Issue
N/A
### Description of changes
- Fix false positive error reported by gosec.


SherinV:
We have lint step here? Does it collide? https://github.com/stolostron/search-v2-operator/blob/6b39184617b8df58163ad8d4431bafea6290dec9/Makefile.prow#L6

PR #175, opened 2024-04-25
ACM-10018 Deploy Global Search resources

jlpadilla:
### Related Issue

https://issues.redhat.com/browse/ACM-10018

### Description of changes
- Validate prerequisites for global search.
  - Global Hub operator deployed in the cluster.
  - Add-on ManagedServiceAccount is enabled.
  - Add-on ClusterProxy is enabled.
- Create or patch resources to configure federated global search. 
  - Patch console-config and concole-mce-config configmaps.
  - Patch search-api deployment
  - Create ManagedService account for each managed hub.
  - Create Manifestwork for each managed hub.


controllers/global_search_setup.go
```diff
@@ -0,0 +1,554 @@
...
+//     b. The ManagedServiceAccount add-on is enabled in the MultiClusterEngine CR.
+//     c. The ClusterProxy addon is enabled in the MultiClusterEngine CR.
+//  3. Enable global search feature in the console.
+//  4. Enable federated search feature in the search-api deployment.
+//  5. Create a ManagedServiceAccount recource for each Managed Hub.
+//  6. Create a ManifestWork resource for each Managed Hub.
+func (r *SearchReconciler) reconcileGlobalSearch(ctx context.Context,
+	instance *searchv1alpha1.Search) (*reconcile.Result, error) {
+
+	if instance.ObjectMeta.Annotations["search.open-cluster-management.io/global-search-peview"] == "true" ||
```
SherinV, line 64:
```suggestion
	if instance.ObjectMeta.Annotations["search.open-cluster-management.io/global-search-preview"] == "true" ||
```

controllers/global_search_setup.go
```diff
@@ -0,0 +1,554 @@
...
+		} else {
+			updateErr := r.updateGlobalSearchStatus(ctx, instance, metav1.Condition{
+				Type:               "GlobalSearchReady",
+				Status:             metav1.ConditionTrue,
+				Reason:             "None",
+				Message:            "None",
+				LastTransitionTime: metav1.Now(),
+			})
+			if updateErr != nil {
+				log.Error(updateErr, "Failed to update the global search status condition on Search CR instance.")
```
SherinV, line 101:
```suggestion
				log.Error(updateErr, "Failed to update the Global Search status condition on Search CR instance.")
```
To be same as above.

controllers/global_search_setup.go
jlpadilla: ahhh, this explains a lot of dev errors.

I added the fully qualified annotation for consistency, but I like `global-search-preview` because it's simpler. Thoughts?

controllers/global_search_setup.go
jlpadilla: done

controllers/global_search_setup.go
SherinV: Shorter and lesser chance of errors. :)

SherinV: Will we remove the preview part for GA?

jlpadilla: My plan for GA is to move it to the spec.

controllers/global_search_setup.go
```diff
@@ -0,0 +1,553 @@
...
+			}
+		}
+	} else {
+		log.Info("The global-search-preview annotation is not present. Checking if global search was enabled before.")
+
+		// Use the status conditions to determine if global search was enabled before this reconcile.
+		globalSearchPresent := false
+		globalSearchConditionIndex := -1
+		for condIndex, condition := range instance.Status.Conditions {
+			if condition.Type == "GlobalSearchReady" {
```
SherinV, line 110: Does it matter if the condition status is _true_ or _false_ here?

jlpadilla: No, it doesn't matter if it's true or false.
The condition "GlobalSearchReady" is added only after global search is enabled. It's not important if it succeeded or failed (true or false) because we need to clean up in both cases. All we need to know is that global search was enabled before.

controllers/global_search_setup.go
```diff
@@ -0,0 +1,553 @@
...
+		log.Error(err, "Failed to enable the federated global search feature.")
+		return err // QUESTION: Should we return here or continue? Should we rollback other changes?
+	}
+
+	// Create configuration resources for each Managed Hub.
+	// MANAGED_HUBS=($(oc get managedcluster -o json | jq -r '.items[] | select(.status.clusterClaims[] |
+	//     .name == "hub.open-cluster-management.io" and .value != "NotInstalled") | .metadata.name'))
+	clusterList, err := r.DynamicClient.Resource(managedClusterResourceGvr).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		log.Error(err, "Failed to list of ManagedClusters to configure global search.")
```
SherinV, line 229:
```suggestion
		log.Error(err, "Failed to list the ManagedClusters to configure global search.")
```

controllers/global_search_setup.go
```diff
@@ -0,0 +1,553 @@
...
+					},
+				},
+			}
+			_, err := r.DynamicClient.Resource(managedServiceAccountGvr).Namespace(cluster.GetName()).
+				Create(ctx, managedSA, metav1.CreateOptions{})
+			if err != nil {
+				log.Error(err, "Failed to create ManagedServiceAccount search-global.")
+				// QUESTION: How should we handle partial errors? Should we continue or rollback?
+			}
+		} else {
```
SherinV, line 280: There might be other error conditions where the client was not able to `Get` the resource - where  !errors.IsNotFound(err).

controllers/global_search_setup.go
```diff
@@ -0,0 +1,553 @@
...
+					},
+					"spec": map[string]interface{}{
+						"rotation": map[string]interface{}{},
+					},
+				},
+			}
+			_, err := r.DynamicClient.Resource(managedServiceAccountGvr).Namespace(cluster.GetName()).
+				Create(ctx, managedSA, metav1.CreateOptions{})
+			if err != nil {
+				log.Error(err, "Failed to create ManagedServiceAccount search-global.")
```
SherinV, line 277:
```suggestion
				log.Error(err, "Failed to create ManagedServiceAccount search-global for ", "cluster ", cluster.GetName())
```

controllers/global_search_setup.go
```diff
@@ -0,0 +1,553 @@
...
+	log.Info("Global search resources configured.")
+	return err
+}
+
+// Create a ManifestWork search-global-config if it doesn't exist.
+func (r *SearchReconciler) createManifestWork(ctx context.Context, cluster string) error {
+	_, err := r.DynamicClient.Resource(manifestWorkGvr).Namespace(cluster).
+		Get(ctx, searchGlobalConfig, metav1.GetOptions{})
+	if err != nil && errors.IsNotFound(err) {
+		log.V(1).Info("CreatingManifestWork search-global-config for Managed Hub", "name", cluster)
```
SherinV, line 300:
```suggestion
		log.V(1).Info("Creating ManifestWork search-global-config for Managed Hub", "name", cluster)
```

controllers/global_search_setup.go
jlpadilla: Here I'm trying to create it only if not found, but maybe it's better if I try to create it and handle the case where it already exists.

controllers/global_search_setup.go
jlpadilla: done

controllers/global_search_setup.go
jlpadilla: I reverted the logic. Now I try to create and ignore the error if already exists.
This is simpler and always 1 API request.

controllers/global_search_setup.go
```diff
@@ -0,0 +1,558 @@
...
+	// Create configuration resources for each Managed Hub.
+	// MANAGED_HUBS=($(oc get managedcluster -o json | jq -r '.items[] | select(.status.clusterClaims[] |
+	//     .name == "hub.open-cluster-management.io" and .value != "NotInstalled") | .metadata.name'))
+	clusterList, err := r.DynamicClient.Resource(managedClusterResourceGvr).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		log.Error(err, "Failed to list the ManagedClusters to configure global search.")
+		return err // QUESTION: Should we return here or continue? Should we rollback other changes?
+	}
+
+	for _, cluster := range clusterList.Items {
```
SherinV, line 233: Put this in `else` or check `if  clusterList is not nil` and items are present? Wondering if it will go into 'nilPointer' error otherwise.

controllers/global_search_setup.go
```diff
@@ -0,0 +1,558 @@
...
+		log.Error(err, "Failed to remove the federated global search feature flag from search-api deployment.")
+	}
+
+	// 3. Delete configuration resources for each Managed Hub.
+	// MANAGED_HUBS=($(oc get managedcluster -o json | jq -r '.items[] | select(.status.clusterClaims[] |
+	//    .name == "hub.open-cluster-management.io" and .value != "NotInstalled") | .metadata.name'))
+	clusterList, err := r.DynamicClient.Resource(managedClusterResourceGvr).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		log.Error(err, "Failed to list ManagedClusters.")
+	}
```
SherinV, line 413: Put this in else or check if  clusterList is not nil and items are present? Wondering if it will go into 'nilPointer' error otherwise.

controllers/global_search_setup.go
```diff
@@ -0,0 +1,558 @@
...
+							},
+							"subjects": []interface{}{
+								map[string]interface{}{
+									"kind":      "ServiceAccount",
+									"name":      searchGlobal,
+									"namespace": "open-cluster-management-agent-addon",
+								},
+							},
+						},
+						// FUTURE: Will remove this Route recource and use cluster-proxy-addon instead.
```
SherinV, line 341:
```suggestion
						// FUTURE: Will remove this Route resource and use cluster-proxy-addon instead.
```

controllers/global_search_setup.go
SherinV: That works great!

controllers/global_search_setup.go
```diff
@@ -0,0 +1,558 @@
...
+// This will create a service account in the Managed Hub and sync the secret containing the access token.
+func (r *SearchReconciler) createManagedServiceAccount(ctx context.Context, cluster string) error {
+	managedSA := &unstructured.Unstructured{
+		Object: map[string]interface{}{
+			"apiVersion": "authentication.open-cluster-management.io/v1beta1",
+			"kind":       "ManagedServiceAccount",
+			"metadata": map[string]interface{}{
+				"name": searchGlobal,
+				"labels": map[string]interface{}{
+					"app": "search",
```
SherinV, line 282: As discussed, adding "feature":"globalSearch" to all the new objects we create might be helpful to see all the things we put in to enable this feature.

SherinV (approved):
Approving - very minor comments for label addition. Thanks, Jorge.

controllers/global_search_setup.go
jlpadilla: done

controllers/global_search_setup.go
jlpadilla: done

controllers/global_search_setup.go
```diff
@@ -0,0 +1,553 @@
...
+						},
+					},
+				},
+			},
+		}
+		_, err := r.DynamicClient.Resource(manifestWorkGvr).Namespace(cluster).
+			Create(ctx, manifestWork, metav1.CreateOptions{})
+		return err
+	}
+	log.V(5).Info("Found existing ManifestWork search-global-config.")
```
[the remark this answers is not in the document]
jlpadilla: done

controllers/global_search_setup.go
jlpadilla: added label `"feature": "global-search"`

controllers/global_search_setup.go
SherinV: FYI @jlpadilla

jlpadilla: Resolved with latest commit. Also updated the delete logic to simplify and reduce api requests.

controllers/global_search_setup.go
```diff
@@ -0,0 +1,570 @@
...
+//	a. MulticlusterGlobalHub operator is installed in the cluster.
+//	b. The ManagedServiceAccount add-on is enabled in the MultiClusterEngine CR.
+//	c. The ClusterProxy addon is enabled in the MultiClusterEngine CR.
+func (r *SearchReconciler) validateGlobalSearchPrerequisites(ctx context.Context) error {
+	log.V(5).Info("Checking global search pre-requisites.")
+	// Verify that MulticlusterGlobalHub operator is installed.
+	// oc get multiclusterglobalhub -A
+	multiclusterGlobalHub, err := r.DynamicClient.Resource(multiclusterGlobalHubGvr).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("MulticlusterGlobalHub operator is not installed.")
```
SherinV, line 156: There can be other errors other than 'Not Found' if the client failed to list. Should we attach the actual error too?

controllers/global_search_setup.go
```diff
@@ -0,0 +1,570 @@
...
+	} else if len(multiclusterGlobalHub.Items) > 0 {
+		log.V(5).Info("Found MulticlusterGlobalHub intance.")
+	}
+
+	// Verify that the ManagedServiceAccount and ClusterProxy add-ons are enabled in the MultiClusterEngine CR.
+	mce, err := r.DynamicClient.Resource(multiclusterengineResourceGvr).
+		Get(ctx, "multiclusterengine", metav1.GetOptions{})
+	if err != nil {
+		log.Error(err, "Failed to get MulticlusterEngine CR.")
+		return fmt.Errorf("Failed to get MulticlusterEngine CR.")
```
SherinV, line 166: Attach original error message too?

controllers/global_search_setup.go
jlpadilla: good point. I wanted to keep it simple because the returned error gets attached to the condition message. What do you think about making it more generic like "Unable to validate the prerequisite MulticlusterGlobalHub operator."
I'll add a log with the full error.

controllers/global_search_setup.go
jlpadilla: The original error message is logged. The message in the return error will get attached to the condition, so I prefer to keep it simple.

controllers/global_search_setup.go
SherinV: Failed to verify if MulticlusterGlobalHub operator is installed, maybe?

SherinV: "Unable to validate the prerequisite MulticlusterGlobalHub operator" sounds better. :)

SherinV (approved):
/lgtm
/hold

PR #179, opened 2024-05-07
Update bundle/manifests

jlpadilla:
### Related Issue
N/A

### Description of changes
- Ran `make bundle` to update files in bundle/manifest.

### DIFF
--- a/bundle/manifests/search-v2-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/search-v2-operator.clusterserviceversion.yaml
@@ -9,13 +9,13 @@ metadata:
           "kind": "Search",
           "metadata": {
             "name": "search-v2-operator"
-            "namespace": "open-cluster-management"
           },
           "spec": null
         }
       ]
     capabilities: Basic Install
-    operators.operatorframework.io/builder: operator-sdk-v1.15.0
+    createdAt: "2024-05-07T20:20:00Z"
+    operators.operatorframework.io/builder: operator-sdk-v1.34.1
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
   name: search-v2-operator.v0.0.1
   namespace: placeholder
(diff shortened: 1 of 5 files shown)


bundle/manifests/search-v2-operator.clusterserviceversion.yaml
```diff
@@ -9,317 +9,230 @@ metadata:
...
+  - base64data: ""
+    mediatype: ""
   install:
     spec:
       clusterPermissions:
-        - rules:
-            - apiGroups:
-                - monitoring.coreos.com
-              resources:
-                - servicemonitors
```
SherinV, line 43: If we don't add these permissions to role.yaml, we will run into errors while trying to create servicemonitors, subjectaccessreviews and such. Will have to reconcile both files.

bundle/manifests/search-v2-operator.clusterserviceversion.yaml
```diff
@@ -9,317 +9,230 @@ metadata:
...
-                        port: 8081
-                      initialDelaySeconds: 15
-                      periodSeconds: 20
-                    name: manager
-                    env:
-                      - name: WATCH_NAMESPACE
-                        valueFrom:
-                          fieldRef:
-                            fieldPath: metadata.namespace
-                      - name: POSTGRES_IMAGE
```
SherinV, line 231: Not sure if these env vars were required. Wondering if they were supposed to be added to some base file.

bundle/manifests/search-v2-operator.clusterserviceversion.yaml
```diff
@@ -9,317 +9,230 @@ metadata:
...
-                - patch
-                - delete
-            - apiGroups:
-                - ""
-              resources:
-                - events
-              verbs:
-                - create
-                - patch
-          serviceAccountName: search-v2-operator
```
SherinV, line 301: My current running operator pod refers this serviceAccountName: search-v2-operator. The updated one shows `        serviceAccountName: search-v2-operator-controller-manager`

bundle/manifests/search-v2-operator.clusterserviceversion.yaml
```diff
@@ -9,317 +9,230 @@ metadata:
...
   installModes:
-    - supported: true
-      type: OwnNamespace
-    - supported: false
-      type: SingleNamespace
-    - supported: false
-      type: MultiNamespace
-    - supported: true
-      type: AllNamespaces
+  - supported: false
```
SherinV, line 218: Have to check if this matters.

bundle/manifests/search-v2-operator.clusterserviceversion.yaml
```diff
@@ -9,317 +9,230 @@ metadata:
...
+  - Advanced Cluster Management
+  - Open Cluster Management
+  - stolostron
+  - acm
+  - ocm
   links:
-    - name: Search Operator
-      url: https://github.com/stolostron/search-v2-operator
+  - name: Search V2 Operator
+    url: https://search-v2-operator.domain
```
SherinV, line 235: We should keep the url? We can update the search-v2-operator/config/manifests/bases/search-v2-operator.clusterserviceversion.yaml so that it reflects here.

bundle/manifests/search-v2-operator.clusterserviceversion.yaml
jlpadilla: So we were adding these manually? I think the correct way would be to add those here. https://github.com/stolostron/search-v2-operator/blob/main/controllers/search_controller.go#L60-L68

Then those should get propagated by `make manifests` and `make bundle`.

bundle/manifests/search-v2-operator.clusterserviceversion.yaml
jlpadilla: changed in `search-v2-operator/config/manifests/bases/search-v2-operator.clusterserviceversion.yaml`

bundle/manifests/search-v2-operator.clusterserviceversion.yaml
jlpadilla: resolved

SherinV (approved):
/lgtm
/hold
Have to check if something breaks when it goes into the build.

bundle/manifests/search-v2-operator.clusterserviceversion.yaml
jlpadilla: Reverted manually. Will fix the automation in separate PR.

bundle/manifests/search-v2-operator.clusterserviceversion.yaml
jlpadilla: Updated the source in config/manifests/bases/search-v2-operator.clusterserviceversion.yaml

bundle/manifests/search-v2-operator.clusterserviceversion.yaml
jlpadilla: Reverted manually, will update automation separately.

PR #186, opened 2024-05-10
Remove wildcard from cluster role verbs

jlpadilla:
### Description of changes
- Remove wildcard from cluster role verbs


SherinV (approved):
/lgtm
/hold

PR #232, opened 2024-10-22
support the customized mce install namespace

hchenxa:
### Related Issue

https://issues.redhat.com/browse/ACM-15075

### Description of changes
- Added ...

### DIFF
--- a/controllers/global_search_setup.go
+++ b/controllers/global_search_setup.go
@@ -215,9 +215,16 @@ func logAndTrackError(errorList *[]error, err error, message string, keysAndValu
 func (r *SearchReconciler) enableGlobalSearch(ctx context.Context, instance *searchv1alpha1.Search) error {
 	errList := []error{} // Using this to allow partial errors and combine at the end.
 
+	mceInstallNS, err := r.getMCETargetNamespace(ctx)
+	if err != nil {
+		logAndTrackError(&errList, err, "Failed to get mce installed namespace.")
+		// will using the default namespace if failed to get the installed namespace from MCE CR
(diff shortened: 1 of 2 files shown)


controllers/global_search_setup.go
```diff
@@ -215,9 +215,11 @@ func logAndTrackError(errorList *[]error, err error, message string, keysAndValu
 func (r *SearchReconciler) enableGlobalSearch(ctx context.Context, instance *searchv1alpha1.Search) error {
 	errList := []error{} // Using this to allow partial errors and combine at the end.
 
+	mceInstallNS, err := r.getMCETargetNamespace(ctx)
+	logAndTrackError(&errList, err, "Failed to get mce installed namespace.")
 	// 1. Enable global search feature in the console.
 	// 1a.Add globalSearchFeatureFlag=true to configmap console-mce-config in multicluster-engine namespace.
-	err := r.updateConsoleConfig(ctx, true, "multicluster-engine", "console-mce-config")
+	err = r.updateConsoleConfig(ctx, true, mceInstallNS, "console-mce-config")
```
jlpadilla, line 222: In case of errors, maybe we should attempt using the default mce namespace.

hchenxa: Thanks @jlpadilla , I think we can check the err and reset to use the default mce namespace to avoid the mceInstallNS is empty

jlpadilla (approved):
/lgtm

controllers/global_search_setup.go
jlpadilla: thanks!

PR #286, opened 2025-02-10
ACM-17312 add search collector service and servicemonitor for prom metrics

smcavey:
### Related Issue

https://issues.redhat.com/browse/ACM-17312

### Description of changes
- Adds search collector service and servicemonitor to scrape prometheus metrics introduced by https://github.com/stolostron/search-collector/pull/400


addon/manifests/chart/templates/service.yaml
```diff
@@ -0,0 +1,29 @@
+---
+{{- if .Values.prometheus.enabled }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "controller.fullname" . }}
```
jlpadilla, line 6: `controller.fullname` doesn't match the template functions defined in `./_helpers.tpl`.  The defined template is called `search.fullname`. You need to make these match.

It's just a name, so we can pick either one. I think I prefer `controller.fullname` because its more generic.
There's also `controller.chart` and `controller.name`. Let's keep al in sync.

addon/manifests/chart/values.yaml
```diff
@@ -23,6 +23,14 @@ logLevel: null
 
 tolerations: []
 
+# This will expose metrics
+prometheus:
+  # This will be automatically enabled if it's an OpenShift cluster
+  enabled: true
```
jlpadilla, line 29: Should we change this back to false?

smcavey: Good question. Now that everything's all ironed out let me set that back to false and wrap the service and servicemonitor around if prometheus.enabled and verify it actually works.

smcavey: https://github.com/stolostron/governance-policy-addon-controller/commit/becfa7e8c034cb2987c981586d7613bc782bbc71#diff-b679eb6ba474be74d700b82fe00aa04d36c5a57e13aae616a14b73fd3a9cd633R109
For the current implementation we have, that comment in our values.yaml is false since our addon code doesn't actually check it.

smcavey: And thus, without some code like that linked, the override didn't happen and the resources weren't created.

smcavey: Updated

addon/addon.go
```diff
@@ -87,7 +93,17 @@ func getValue(cluster *clusterv1.ManagedCluster,
...
+	}
+	for _, cc := range cluster.Status.ClusterClaims {
+		if cc.Name == "product.open-cluster-management.io" {
+			addonValues.KubernetesDistribution = cc.Value
+			break
+		}
 	}
+	// enable prometheus if on OpenShift
+	addonValues.Prometheus["enabled"] = addonValues.KubernetesDistribution == "OpenShift"
+
```
jlpadilla, line 106: Why do we need this?

addon/manifests/chart/templates/deployment.yaml
```diff
@@ -84,8 +84,21 @@ spec:
           limits:
             memory: 1024Mi
           {{- end }}
+        {{- if .Values.prometheus.enabled }}
```
jlpadilla, line 87: I don't think we need this if condition.
liveness and readiness endpoints are provided by the image. It is independent of prometheus.

addon/manifests/chart/templates/service.yaml
```diff
@@ -0,0 +1,26 @@
+---
+{{- if .Values.prometheus.enabled }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ template "controller.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ template "controller.fullname" . }}
+    chart: {{ template "search.chart" . }}
```
jlpadilla, line 10: Maybe this should be `controller.chart` for consistency. Also need to update in `_helpers.tpl`

addon/addon.go
smcavey: We need this to flip .Values.prometheus.enabled to true if the managed cluster is OpenShift. Else defaulting .Values.prometheus.enabled to false, the service and servicemonitor will never get created on the managed cluster

addon/manifests/chart/templates/service.yaml
```diff
@@ -0,0 +1,26 @@
...
+    release: {{ .Release.Name }}
+    search-monitor: {{ template "controller.fullname" . }}
+    heritage: {{ .Release.Service }}
+    addon.open-cluster-management.io/hosted-manifest-location: hosting
```
jlpadilla, line 14: I don't think this label is needed. I'm not sure what it does and we don't have it on the existing resources in this chart.

addon/manifests/chart/templates/service_monitor.yaml
```diff
@@ -0,0 +1,25 @@
...
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: {{ template "controller.fullname" . }}-monitor
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ template "controller.fullname" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+    addon.open-cluster-management.io/hosted-manifest-location: hosting
```
jlpadilla, line 12: - Remove `addon.open-cluster-management.io/hosted-manifest-location: hosting`
- Add `component: search`

jlpadilla (approved):
/lgtm

PR #325, opened 2025-03-11
ACM-18716 Update to go1.22

jlpadilla:
### Related Issue
https://issues.redhat.com/browse/ACM-18716

### Description of changes
- Update go to 1.22
- Update golangci-lint to 1.64.6

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # Copyright Contributors to the Open Cluster Management project
 
-FROM registry.ci.openshift.org/stolostron/builder:go1.21-linux AS builder
+FROM registry.ci.openshift.org/stolostron/builder:go1.22-linux AS builder
 
 WORKDIR /workspace
 # Copy the Go Modules manifests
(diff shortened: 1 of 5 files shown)


smcavey (approved):

Dockerfile
```diff
@@ -1,6 +1,6 @@
 # Copyright Contributors to the Open Cluster Management project
 
-FROM registry.ci.openshift.org/stolostron/builder:go1.21-linux AS builder
+FROM registry.ci.openshift.org/stolostron/builder:go1.22-linux AS builder
```
dhaiducek, line 3: @jlpadilla The base image here is still UBI8. The 1.22 builder image is UBI9, so the build will work but the binary won't run.

In addition, you'll need to update the Prow config to match:
https://github.com/openshift/release/blob/d12092f0d8a95f83a89809d03187ea2b70367743/ci-operator/config/stolostron/search-v2-operator/stolostron-search-v2-operator-main.yaml#L5

jlpadilla: @dhaiducek Thank you for all your help with prow and konflux!

Updating with this PR. https://github.com/openshift/release/pull/62742

PR #363, opened 2025-03-29
2.12 fix cves

smcavey:
fix cves from https://github.com/stolostron/search-v2-operator/pull/352


jlpadilla:
Merging manually to override konflux enterprise contract job error.

PR #525, opened 2025-05-21
ACM-19916: Add fine-grained rbac tech preview config

jlpadilla:
### Related Issue

https://issues.redhat.com/browse/ACM-19916

### Description of changes
- Use the annotation `fine-grained-rbac-preview` to enable the fine-grained rbac feature in the search-api.

### DIFF
--- a/controllers/fine_grained_rbac.go
+++ b/controllers/fine_grained_rbac.go
@@ -25,7 +25,7 @@ func (r *SearchReconciler) reconcileFineGrainedRBACConfiguration(ctx context.Con
 			corev1.EnvVar{Name: "FEATURE_FINE_GRAINED_RBAC", Value: "true"})
 		if err != nil {
 			log.Error(err, "Failed to configure fine-grained RBAC.")
-			r.updateVMStatus(ctx, instance, metav1.Condition{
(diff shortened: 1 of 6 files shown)


controllers/fine_grained_rbac.go
```diff
@@ -0,0 +1,78 @@
...
+				Type:               CONDITION_FINE_GRAINED_RBAC,
+				Status:             metav1.ConditionFalse,
+				Reason:             "ConfigurationError",
+				Message:            "Failed to enable fine-grained RBAC.",
+				LastTransitionTime: metav1.Now(),
+			})
+			return &reconcile.Result{}, err
+		}
+
+		r.updateVMStatus(ctx, instance, metav1.Condition{
```
smcavey, line 38: Think we should generalize this name now

smcavey:
Tested it all out in my env. Works as expected. Added annotation and observed the env variable get updated in the search api and status condition in search, set annotation to false and observed the env variable get unset in the search api and status condition removed in search

controllers/fine_grained_rbac.go
jlpadilla: good point. changed.

PR #537, opened 2025-05-27
ACM-19020 - support addondeploymentconfig resourcerequirements config

smcavey:
### Related Issue

https://issues.redhat.com/browse/ACM-19020

### Description of changes
- Support configurable resource requests and limits using the dev guidelines: https://github.com/stolostron/foundation-docs/blob/main/addon/addon-resource-requirements.md

This change continues to permit resource configuration via the annotations to avoid breaking existing user experience. The priority is annotations > resourcerequirements > default. If there's no annotations, we use resourcerequirements, if there's neither, we use the existing defaults.

### DIFF
--- a/addon/manifests/chart/templates/deployment.yaml
+++ b/addon/manifests/chart/templates/deployment.yaml
@@ -69,30 +69,6 @@ spec:
         - name: NO_PROXY
           value: {{ .Values.global.proxyConfig.NO_PROXY }}
         {{- end }}
-#        resources:
-#          {{- $memLimit := .Values.userargs.limitMemory }}
-#          {{- $memRequest := .Values.userargs.requestMemory }}
-#          {{- $rrs := .Values.global.resourceRequirements | default list }}
-#          {{- $match := dict }}
-#          {{- range $rrs }}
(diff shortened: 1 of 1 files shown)


addon/manifests/chart/templates/deployment.yaml
```diff
@@ -70,17 +70,25 @@ spec:
...
+          {{- $memLimit := .Values.userargs.limitMemory }}
+          {{- $memRequest := .Values.userargs.requestMemory }}
+          {{- $rrs := .Values.global.resourceRequirements | default list }}
+          {{- $match := dict }}
+          {{- range $rrs }}
+            {{- if regexMatch .containerIDRegex "deployments:klusterlet-addon-search:collector" }}
+              {{- $match = .resources }}
+            {{- end }}
+          {{- end }}
+          {{- if and $memLimit $memRequest }}
```
zlayne, line 82: Could users set only one of the memory values?

smcavey: Not in its present form. I could take the values from annotations, if either are missing, attempt to fill in the blanks from the resourcerequirements, and if there's still one missing, fill it in with the default?

zlayne (approved):

PR #538, opened 2025-05-29
manual update imdario mergo dependency

smcavey:
manually updating imdario/mergo that https://github.com/stolostron/search-v2-operator/pull/536 tried to do. imdario/mergo changed its path to dario.cat/mergo so we have to replace old path reference


jlpadilla (approved):
/lgtm

PR #551, opened 2025-06-09
main security updates

smcavey:
bumping golang.org/x/net and google.golang.org/protobuf for security reasons. Bumping versions to be in line with other repos


jlpadilla (approved):
/lgtm

PR #638, opened 2025-08-19
ACM-22418 Set default postgres statement_timeout

jlpadilla:
### Related Issue
https://issues.redhat.com/browse/ACM-22418

### Description of changes
- Sets postgres default `statement_timeout = '60000'` to provide a fail-safe to limit the impact from inefficient queries.

### DIFF
--- a/controllers/create_pgconfigmap.go
+++ b/controllers/create_pgconfigmap.go
@@ -26,7 +26,8 @@ func (r *SearchReconciler) PostgresConfigmap(instance *searchv1alpha1.Search) *c
 ssl_cert_file = '/sslcert/tls.crt'
 ssl_key_file = '/sslcert/tls.key'
 ssl_ciphers = 'HIGH:!aNULL'
-max_parallel_workers_per_gather = '8'`
+max_parallel_workers_per_gather = '8'
+statement_timeout = '60000'`
 
 	data[startScript] = `psql -d search -U searchuser -c "CREATE SCHEMA IF NOT EXISTS search"
 psql -d search -U searchuser -c "CREATE TABLE IF NOT EXISTS search.resources (uid TEXT PRIMARY KEY, cluster TEXT, data JSONB)"


jlpadilla:
/hold
This PR is missing the upgrade scenario.

smcavey:
Overall I like this proposal and that it keeps the default config preserved above the custom configuration separated by the comment `# Customizations ...`. I don't see any downside since this is already a resource that users could already edit. Regarding the upgrade case, I suppose the very first check we need to make in the special handling block if check if `found.Data["postgresql.conf"] != cm.Data["postgresql.conf"]` and push the found config onto the custom config so we don't change users' existing configs?

controllers/common.go
```diff
@@ -401,6 +402,50 @@ func (r *SearchReconciler) createConfigMap(ctx context.Context, cm *corev1.Confi
...
+			// Check if update is needed
+			if !equality.Semantic.DeepEqual(found.Data, cm.Data) {
+				log.Info("Update is needed for configmap ", "name", cm.Name)
+
+				err = r.Update(ctx, cm)
+				if err != nil {
+					log.Error(err, "Could not update configmap")
+					return &reconcile.Result{}, err
+				}
+				log.Info("Updated configmap ", "name", cm.Name)
```
smcavey, line 441: Maybe log here or additionally that a restart is required to take effect?

jlpadilla: done

smcavey:
Looks good:
`oc get configmap search-postgres -o yaml
apiVersion: v1
data:
  custom-postgresql.conf: |
    # Customizations appended to postgresql.conf.
    statement_timeout = '70000'
  postgresql-start.sh: |-
...
  postgresql.conf: |
...
    statement_timeout = '60000'
    # Customizations appended to postgresql.conf.
    statement_timeout = '70000'
`
And then removing the statement_timeout from the custom config reverts it all back to initial state. And log to restart Postgres observed both times :+1:

smcavey:
I'm satisfied with the tests. In running I also overrode the statement_timeout in my pre-existing config from 60000 -> 70000 and after migrating it moved the 70000 to custom config. Another test added a new config log_connections to the config, updated the operator image, and it applied it also properly.
/lgtm

PR #652, opened 2025-08-26
[release-2.13] fix mutli-arch builds

smcavey:
### Description of changes
- Fixing multi-arch builds. all non amd64 build-images jobs kept failing because for all other arch types, amd64 was hard-coded in.


jlpadilla (approved):
/lgtm

PR #706, opened 2026-04-15
add fieldSuffix to collectorconfig to prevent collisions

smcavey:
### Description of changes
- Added fieldSuffix to collectorConfig to avoid collisions with built in field collection

## Summary by CodeRabbit

* **New Features**
  * Collection rules gain a configurable field suffix: append a custom suffix to indexed field names to avoid collisions with built-in fields; built-in fields continue to take precedence on name conflicts.
* **Chores**
  * Internal tooling and lint-suppression annotations added to service and monitoring configurations (no behavioral changes).

### DIFF
--- a/api/v1alpha1/collectorconfig_types.go
+++ b/api/v1alpha1/collectorconfig_types.go
@@ -99,9 +99,9 @@ type CollectionRule struct {
 	// +optional
 	// +kubebuilder:default=""
 	// FieldSuffix appended to custom field names in this rule to avoid collisions with built-in fields.
-	// For example, if fieldSuffix is "_grc", a field named "status" becomes "status_grc".
+	// For example, if fieldSuffix is "grc", a field named "status" becomes "status.grc".
 	// Defaults to "" (no suffix). If a collision is detected, the built-in field takes precedence.
-	// This allows different teams (GRC, Virt, etc.) to use different suffixes in the same config.
+	// This allows different extensions (GRC, Virt, etc.) to use different suffixes in the same config.
 	FieldSuffix string `json:"fieldSuffix,omitempty"`
 
 	// +optional
--- a/api/v1alpha1/collectorconfig_types.go
+++ b/api/v1alpha1/collectorconfig_types.go
@@ -97,6 +97,14 @@ type CollectionRule struct {
 	Fields []Field `json:"fields,omitempty"`
 
 	// +optional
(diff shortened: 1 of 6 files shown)


api/v1alpha1/collectorconfig_types.go
```diff
@@ -96,6 +96,14 @@ type CollectionRule struct {
 	// Specifies the additional fields on the resource to index by Search Collectors.
 	Fields []Field `json:"fields,omitempty"`
 
+	// +optional
+	// +kubebuilder:default=""
+	// FieldSuffix appended to custom field names in this rule to avoid collisions with built-in fields.
+	// For example, if fieldSuffix is "_grc", a field named "status" becomes "status_grc".
+	// Defaults to "" (no suffix). If a collision is detected, the built-in field takes precedence.
+	// This allows different teams (GRC, Virt, etc.) to use different suffixes in the same config.
```
jlpadilla: Change "teams" to extensions.
```suggestion
	// This allows different extensions (GRC, Virt, etc.) to use different suffixes in the same config.
```

api/v1alpha1/collectorconfig_types.go
```diff
@@ -96,6 +96,14 @@ type CollectionRule struct {
 	// Specifies the additional fields on the resource to index by Search Collectors.
 	Fields []Field `json:"fields,omitempty"`
 
+	// +optional
+	// +kubebuilder:default=""
+	// FieldSuffix appended to custom field names in this rule to avoid collisions with built-in fields.
+	// For example, if fieldSuffix is "_grc", a field named "status" becomes "status_grc".
```
jlpadilla: Should the suffix include the separator `_grc` ?
I expect it to be just `grc` and we control the separator character.

Also we should probably use a different character like `.` and prevent users from using the character in any property name. The `_` already has a special meaning.

smcavey: Probably simpler we add the separator. Updated this comment `_grc` -> `grc` and `status_grc` -> `status.grc`, users can infer we add the `.` or we can omit the separator from the description and leave the wording that we will suffix it and they'll figure it out

jlpadilla: I'm good with the latest update. Thanks

jlpadilla (approved):
/lgtm

PR #727, opened 2026-05-19
ACM-2071 - CollectorConfig enable collectConditions and validate

smcavey:
### Related Issue

https://issues.redhat.com/browse/ACM-2071

### Description of changes
- Remove not implemented description for collectConditions. Permits specifying kinds = ["*"] for apigroups when trying to collect status conditions apigroup wide.

## Summary by CodeRabbit

* **New Features**
  * Documented CollectConditions behavior and clarified wildcard ("*") semantics for kinds.

* **Improvements**
  * Validation updated to disallow wildcard kinds when fields are specified and to enforce CollectConditions constraints.

* **Tests**
  * Added validation tests covering CollectConditions, wildcard kinds, and API group requirements.

* **Documentation**
  * CRD schema descriptions updated to reflect the above behaviors.

### DIFF
--- a/api/v1alpha1/collectorconfig_types.go
+++ b/api/v1alpha1/collectorconfig_types.go
@@ -114,7 +114,7 @@ type CollectionRule struct {
 	CollectAnnotations *bool `json:"collectAnnotations,omitempty"`
 
 	// +optional
-	// Specifies to collect status conditions for the resource.
+	// Enable collection of status conditions for the resource.
 	CollectConditions *bool `json:"collectConditions,omitempty"`
 
 	// +optional
@@ -127,7 +127,7 @@ type ResourceSelector struct {
 	// Specifies apiGroups of resources.
 	APIGroups []string `json:"apiGroups"`
 
-	// Specifies kinds of resources. Use "*" to match all kinds in the apiGroup (only with collectConditions).
+	// Specifies kinds of resources. Use "*" to match all kinds in the apiGroup (not permitted with fields).
 	Kinds []string `json:"kinds"`
 }
 
--- a/api/v1alpha1/collectorconfig_types.go
+++ b/api/v1alpha1/collectorconfig_types.go
@@ -114,7 +114,7 @@ type CollectionRule struct {
 	CollectAnnotations *bool `json:"collectAnnotations,omitempty"`
 
 	// +optional
-	// [NOT IMPLEMENTED] Specifies to collect status conditions for the resource.
+	// Specifies to collect status conditions for the resource.
 	CollectConditions *bool `json:"collectConditions,omitempty"`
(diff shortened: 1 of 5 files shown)


api/v1alpha1/collectorconfig_webhook.go
```diff
@@ -96,8 +96,31 @@ func (r *CollectorConfig) validateCollectorConfig() error {
...
+		hasCollectConditions := rule.CollectConditions != nil && *rule.CollectConditions
+		hasFields := len(rule.Fields) > 0
+
+		// Validate wildcard kind "*": only allowed with collectConditions and without fields
+		for _, k := range rule.ResourceSelector.Kinds {
+			if k == "*" && !hasCollectConditions {
+				allErrs = append(allErrs, field.Invalid(
+					rulePath.Child("resourceSelector", "kinds"),
+					rule.ResourceSelector.Kinds,
+					"wildcard kind \"*\" is only allowed when collectConditions is true",
```
jlpadilla: Will this extend to other things like annotations and printerColumns? Maybe this check and message is too specific.

api/v1alpha1/collectorconfig_types.go
```diff
@@ -114,7 +114,7 @@ type CollectionRule struct {
 	CollectAnnotations *bool `json:"collectAnnotations,omitempty"`
 
 	// +optional
-	// [NOT IMPLEMENTED] Specifies to collect status conditions for the resource.
+	// Specifies to collect status conditions for the resource.
```
jlpadilla: Does this sound better?
```suggestion
	// Enable collection of status conditions for the resource.
```

api/v1alpha1/collectorconfig_webhook.go
smcavey: This check here `if k == "*" && !hasCollectConditions` could be removed entirely to generalize for annotation, conditions, printerColumns, and specifying which resources to include/exclude from collection. Just checking `if k == "*" && hasFields` should suffice

smcavey: Updated. Should be more general. Only checks that a wildcard kinds isn't provided only when there are also fields present.

api/v1alpha1/collectorconfig_types.go
smcavey: Updated this comment and the kinds description to say wildcard kinds is not permitted with fields

api/v1alpha1/collectorconfig_webhook.go
jlpadilla: Code is also simpler, so double win! 💯

jlpadilla (approved):
/lgtm

PR #729, opened 2026-05-29
ACM-21883 - integration collector config

smcavey:
### Related Issue

https://issues.redhat.com/browse/ACM-21883

### Description of changes
- Added integration teams CollectorConfig and process for merging customer + integration team into a meged CollectorConfig to be finally consumed by the collector

## Summary by CodeRabbit

* **New Features**
  * Operator-managed collector configs are protected from unauthorized create/update/delete.
  * Webhook validation now runs on deletes as well as creates/updates.
  * Collector configs from multiple sources are automatically merged deterministically; user configs can augment merged output.

* **Chores**
  * Expanded RBAC to support collector configuration management and webhook operations.

* **Tests**
  * Added tests for admission protection, merge logic, reconciliation, and related edge cases.

### DIFF
--- a/api/v1alpha1/collectorconfig_webhook.go
+++ b/api/v1alpha1/collectorconfig_webhook.go
@@ -17,10 +17,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
 
-const (
-	operatorServiceAccount = "search-serviceaccount"
-)
-
 // log is for logging in this package.
 var collectorconfiglog = logf.Log.WithName("collectorconfig-resource")
 
@@ -69,7 +65,7 @@ func (r *CollectorConfig) ValidateUpdate(ctx context.Context, oldObj, newObj run
 	}
 	collectorconfiglog.Info("validate update", "name", cc.Name)
 
-	// Check both old and new objects — prevents stripping the integration label to bypass protection.
+	// Check both old and new objects — prevents stripping the owner reference to bypass protection.
 	if err := rejectIfProtected(ctx, cc); err != nil {
 		return nil, err
 	}
@@ -285,33 +281,36 @@ func contains(slice []string, item string) bool {
 	return false
 }
 
-// isProtectedConfig returns true for configs managed by the operator.
-// This includes merged-collector-config (by name) and any config labeled as an integration team config.
-func isProtectedConfig(name string, labels map[string]string) bool {
-	if name == "merged-collector-config" {
-		return true
-	}
-	if labels[IntegrationTeamLabel] == IntegrationTeamLabelValue {
-		return true
+// isOperatorOwned returns true if the CollectorConfig has a controller owner reference,
+// indicating it is managed by the operator and should not be modified directly.
+func isOperatorOwned(cc *CollectorConfig) bool {
+	for _, ref := range cc.GetOwnerReferences() {
+		if ref.Controller != nil && *ref.Controller {
+			return true
+		}
 	}
 	return false
 }
 
-// isOperatorServiceAccount checks whether the request originates from the operator's service account
-// in the same namespace as the CollectorConfig being acted on.
-func isOperatorServiceAccount(ctx context.Context) bool {
+// rejectIfProtected returns an error if the config has a controller owner reference
+// and the caller is not a service account in the same namespace (i.e., not the operator).
+func rejectIfProtected(ctx context.Context, cc *CollectorConfig) error {
+	if !isOperatorOwned(cc) {
+		return nil
+	}
+
 	req, err := admission.RequestFromContext(ctx)
 	if err != nil {
-		return false
+		return fmt.Errorf("could not extract admission request: %v", err)
 	}
-	expectedUser := fmt.Sprintf("system:serviceaccount:%s:%s", req.Namespace, operatorServiceAccount)
-	return req.UserInfo.Username == expectedUser
-}
 
-// rejectIfProtected returns an error if the config is operator-managed and the caller is not the operator SA.
-func rejectIfProtected(ctx context.Context, cc *CollectorConfig) error {
-	if isProtectedConfig(cc.Name, cc.Labels) && !isOperatorServiceAccount(ctx) {
-		return fmt.Errorf("%s is managed by the search operator and cannot be modified directly", cc.Name)
+	// Allow any service account in the resource's namespace.
+	// Only the operator namespace should have write RBAC for CollectorConfigs,
+	// so namespace-scoped SA check is sufficient.
+	expectedPrefix := fmt.Sprintf("system:serviceaccount:%s:", req.Namespace)
+	if strings.HasPrefix(req.UserInfo.Username, expectedPrefix) {
+		return nil
 	}
-	return nil
+
+	return fmt.Errorf("%s is managed by the search operator and cannot be modified directly", cc.Name)
 }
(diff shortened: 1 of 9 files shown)


controllers/create_collectorconfig.go
```diff
@@ -0,0 +1,180 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	customerCollectorConfigName    = "customer-collector-config"
+	integrationCollectorConfigName = "integration-collector-config"
```
jlpadilla: We may have multiple integration configurations, not sure if this naming will allow for that.

controllers/search_controller.go
```diff
@@ -405,6 +416,28 @@ func (r *SearchReconciler) SetupWithManager(mgr ctrl.Manager) error {
 				return nil
 			}),
 		).
+		Watches(&searchv1alpha1.CollectorConfig{}, handler.EnqueueRequestsFromMapFunc(
+			func(ctx context.Context, a client.Object) []reconcile.Request {
+				name := a.GetName()
+				// Skip operator-managed outputs to prevent reconcile loops.
+				if name == mergedCollectorConfigName || name == integrationCollectorConfigName {
+					return nil
+				}
```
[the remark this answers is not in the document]
smcavey: The integration-collector-config and merged-collector-config shouldn't ever be deleted. They may be empty, but they shouldn't ever not exist. They are a product of other configs and protected resources, see `rejectIfProtected()` in `collectorconfig_webhook.go`

controllers/create_collectorconfig.go
```diff
@@ -0,0 +1,121 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	customerCollectorConfigName = "customer-collector-config"
```
jlpadilla: Wondering if "user-collector-config" is better than "customer". Thinking in the context of open source usage.

jlpadilla: Also wondering if we should allow customer to define multiple CollectorConfigs. In that case, a label is better than a fixed name.

smcavey: Renamed. Also capturing here our conversation about the edge case that arises when we would try to merge multiple user config namespaceSelectors

api/v1alpha1/collectorconfig_webhook.go
```diff
@@ -261,3 +284,33 @@ func contains(slice []string, item string) bool {
 	}
 	return false
 }
+
+// isProtectedConfig returns true for configs managed by the operator.
+// This includes merged-collector-config (by name) and any config labeled as an integration team config.
+func isProtectedConfig(name string, labels map[string]string) bool {
```
jlpadilla: I wonder if it's better to use the owner field instead. If the CollectorConfig is owned by the operator, then it's protected.

Also thinking if we should provide a special label to override this protection. The scenario I'm thinking is if CNV hits a critical bug that requires a change to their config we could provide a workaround using the override. Otherwise, the only way to make a change would be publishing a new version of the operator.

api/v1alpha1/collectorconfig_webhook.go
```diff
@@ -261,3 +284,33 @@ func contains(slice []string, item string) bool {
...
+	return false
+}
+
+// isOperatorServiceAccount checks whether the request originates from the operator's service account.
+func isOperatorServiceAccount(ctx context.Context) bool {
+	req, err := admission.RequestFromContext(ctx)
+	if err != nil {
+		return false
+	}
+	return strings.HasPrefix(req.UserInfo.Username, "system:serviceaccount:") &&
```
jlpadilla: 🤖 AI Review
**Security: `isOperatorServiceAccount` doesn't verify the SA's namespace.**

The current check:
```go
strings.HasPrefix(req.UserInfo.Username, "system:serviceaccount:") &&
strings.HasSuffix(req.UserInfo.Username, ":"+operatorServiceAccount)
```

passes for `system:serviceaccount:attacker-ns:search-serviceaccount` — any SA named `search-serviceaccount` in any namespace bypasses the protection. In a multi-tenant cluster where different teams can create SAs, this is a meaningful attack surface.

The username format is `system:serviceaccount:<namespace>:<sa-name>`. The check should also verify the namespace. Since the namespace is runtime-determined (`POD_NAMESPACE` env var), consider building the expected full username:
```go
expectedUser := fmt.Sprintf("system:serviceaccount:%s:%s", operatorNamespace, operatorServiceAccount)
return req.UserInfo.Username == expectedUser
```

No existing test covers a `search-serviceaccount` SA in a non-operator namespace.

controllers/create_collectorconfig.go
```diff
@@ -0,0 +1,121 @@
...
+				APIVersion: searchv1alpha1.GroupVersion.String(),
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      mergedCollectorConfigName,
+				Namespace: namespace,
+			},
+			Spec: mergedSpec,
+		}
+		if errRef := controllerutil.SetControllerReference(instance, merged, r.Scheme); errRef != nil {
+			log.V(2).Info("Could not set controller reference for merged-collector-config")
```
jlpadilla: 🤖 AI Review

**`SetControllerReference` error is still silently swallowed.**

CodeRabbit flagged this and it was marked "✅ Addressed", but the current HEAD still has:
```go
if errRef := controllerutil.SetControllerReference(instance, merged, r.Scheme); errRef != nil {
    log.V(2).Info("Could not set controller reference for merged-collector-config")
}
```

If `SetControllerReference` fails, `merged-collector-config` is still created — without an owner reference. It then won't be garbage-collected when the `Search` CR is deleted.

Suggested fix:
```go
if errRef := controllerutil.SetControllerReference(instance, merged, r.Scheme); errRef != nil {
    log.Error(errRef, "Could not set controller reference for merged-collector-config")
    return &reconcile.Result{}, errRef
}
```

controllers/create_collectorconfig.go
jlpadilla, line 1: 🤖 AI Review

**Design question: how do integration teams deploy their labeled configs?**

The operator discovers integration configs by label (`search.open-cluster-management.io/config-type: integration`), but the webhook blocks any non-operator SA from creating a config with that label. This creates a circular dependency:

- CNV/governance operator SA tries to create a labeled config → **webhook rejects it**
- Without the label, the operator can't discover it → **rules never merged**

If integration configs are always bundled inside the search operator itself (i.e., the operator creates them as part of its own reconciliation), that's fine — but then the label-based discovery is just for operator-created resources, not for external teams self-registering.

What's the intended deployment model? Can integration teams (CNV, governance) ship their own `CollectorConfig` CRs independently, or do they always go through the search operator bundle?

jlpadilla (changes requested):
Two concrete fixes needed before merge — see inline comments. Also left a design question about how integration teams deploy their configs.

jlpadilla (approved):
/lgtm

PR #730, opened 2026-06-01
ACM-34617 - add postgres lifecycle preStop hook

smcavey:
### Description of changes
- Added lifecycle preStop hook per https://redhat-internal.slack.com/archives/C05KPMFKDFX/p1780061585986619
- https://redhat.atlassian.net/browse/ACM-34617

## Summary by CodeRabbit

* **Chores**
  * Enhanced graceful shutdown of PostgreSQL during pod termination with updated timeout configuration.

### DIFF
--- a/controllers/create_pgdeployment.go
+++ b/controllers/create_pgdeployment.go
@@ -73,6 +73,17 @@ func (r *SearchReconciler) PGDeployment(instance *searchv1alpha1.Search) *appsv1
 				},
 			},
 		},
+		Lifecycle: &corev1.Lifecycle{
+			PreStop: &corev1.LifecycleHandler{
+				Exec: &corev1.ExecAction{
+					Command: []string{
+						"/bin/sh", "-c",
+						// disconnect clients, wait for server shutdown, timeout after 55 seconds leaving 5-second terminationGracePeriod
+						"pg_ctl stop -D /var/lib/pgsql/data/userdata -m fast -w -t 55",
+					},
+				},
+			},
+		},
 	}
 	args := getContainerArgs(deploymentName, instance)
 	if args != nil {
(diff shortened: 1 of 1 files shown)


controllers/create_pgdeployment.go
```diff
@@ -73,6 +73,17 @@ func (r *SearchReconciler) PGDeployment(instance *searchv1alpha1.Search) *appsv1
 				},
 			},
 		},
+		Lifecycle: &corev1.Lifecycle{
+			PreStop: &corev1.LifecycleHandler{
+				Exec: &corev1.ExecAction{
+					Command: []string{
+						"/bin/sh", "-c",
+						// disconnect clients, wait for server shutdown, timeout after 55 seconds leaving 5-second terminationGracePeriod
+						"pg_ctl stop -D /var/lib/pgsql/data/userdata -m fast -w -t 55",
```
jlpadilla, line 82: I'm wondering if fast mode could cause problems in indexer or api components because it won't wait for clients to disconnect.

https://www.postgresql.org/docs/current/app-pg-ctl.html#:~:text=%E2%80%9CSmart%E2%80%9D,is%20shut%20down.
> “Smart” mode disallows new connections, then waits for all existing clients to disconnect. If the server is in hot standby, recovery and streaming replication will be terminated once all clients have disconnected. “Fast” mode (the default) does not wait for clients to disconnect. All active transactions are rolled back and clients are forcibly disconnected, then the server is shut down.

jlpadilla: Did some research and decided that this is our best alternative.

The only way to handle the disconnect better would be with a proxy layer like PgBouncer. Currently we don't have this and it would add complexity to our configuration.

> If you are using a dedicated proxy layer like PgBouncer, you can safely orchestrate a zero-downtime or graceful server termination from the pooler side instead of the database side:
> 1. PAUSE Command: You issue a PAUSE command to PgBouncer. PgBouncer will let active transactions finish completely, but it will buffer and queue any new incoming client queries.
> 2. Server Maintenance: Once active transactions drain, you can safely restart or terminate the PostgreSQL server.
> 3. RESUME Command: After PostgreSQL boots back up, you issue a RESUME command to PgBouncer. The queued application queries are then forwarded to the new server without throwing errors to your users.

jlpadilla (approved):
/lgtm

PR #738, opened 2026-06-16
ACM-20052: Ensure CollectorConfig resources are included in ACM hub backups

sahare:
## Summary

Fixes ACM-20052: CollectorConfig CRs created by users are currently not backed up during ACM hub backup/restore because the entire `search.open-cluster-management.io` API group is in the backup operator's `excludedAPIGroups` list. Resources in excluded groups are only included in the `acm-resources-generic-schedule` backup when they carry the label `cluster.open-cluster-management.io/backup`.

This PR adds a reconciler step that automatically labels all user-managed CollectorConfig CRs in the operator namespace, ensuring they survive a hub backup/restore cycle without any manual action from the user.

## Changes

**`api/v1alpha1/collectorconfig_types.go`**
- Add `BackupLabel = "cluster.open-cluster-management.io/backup"` constant

**`controllers/create_collectorconfig.go`**
- Add `ensureCollectorConfigsBackupLabel` — lists all CollectorConfigs in the namespace, skips `merged-collector-config` (operator-managed, rebuilt on every reconcile), patches the backup label onto any config missing it

**`controllers/search_controller.go`**
- Call `ensureCollectorConfigsBackupLabel` in the reconcile loop after `createOrUpdateMergedCollectorConfig`

**`controllers/create_collectorconfig_test.go`**
- `TestBackupLabel_UserConfig_GetsLabeled` — user config without label gets labeled
- `TestBackupLabel_IntegrationTeamConfig_GetsLabeled` — integration team config gets labeled
- `TestBackupLabel_MergedConfig_Skipped` — merged-collector-config is NOT labeled
- `TestBackupLabel_AlreadyLabeled_NotPatched` — no patch issued if label already present
- `TestBackupLabel_MultipleConfigs` — mixed scenario: user + team labeled, merged skipped

## Design notes

- `merged-collector-config` is excluded because it is entirely derived from the user and integration team configs; backing it up would restore a stale merged view instead of the source configs, which the operator re-merges correctly on first reconcile after restore.
- The label value is `""` (empty string), which is the ACM backup convention for inclusion in `acm-resources-generic-schedule`.
- Spoke-side CollectorConfig (on managed clusters) is out of scope for this Jira; those configs are not hub resources and require a separate distribution mechanism (future work).

## Test plan

- [x] `go test ./...` — all tests pass
- [x] `golangci-lint run` — 0 issues
- [x] E2E on ACM 5.0.0-111 cluster:
  1. Deployed custom operator — `user-collector-config` labeled within ~20s of creation
  2. `merged-collector-config` confirmed unlabeled
  3. Velero backup confirmed `user-collector-config` in resource list under `search.open-cluster-management.io/v1alpha1/CollectorConfig`
  4. Deleted `user-collector-config`, ran Velero restore — resource restored with full spec intact

## Summary by CodeRabbit

* **New Features**
  * Collector configurations are now automatically prepared for backup/restore during reconciliation by applying a backup label.
  * User-managed configurations receive the label; operator-managed merged configurations are excluded.
  * Labeled configurations are left unchanged to avoid unnecessary updates.
* **Bug Fixes**
  * Backup label handling is now consistent across collector-related ConfigMaps, including update behavior when the label is missing.
* **Tests**
  * Added unit tests covering user and integration-team labeling, merged-config skip behavior, no-repatching for already-labeled resources, and mixed multi-config cases.

### DIFF
--- a/api/v1alpha1/collectorconfig_types.go
+++ b/api/v1alpha1/collectorconfig_types.go
@@ -40,6 +40,11 @@ const IntegrationTeamLabel = "search.open-cluster-management.io/config-type"
 // IntegrationTeamLabelValue is the expected value for IntegrationTeamLabel.
 const IntegrationTeamLabelValue = "integration"
 
+// BackupLabel is the label applied to CollectorConfig CRs so they are included in ACM hub backups.
+// The search.open-cluster-management.io API group is excluded from the automatic resources backup,
+// but resources carrying this label are picked up by the acm-resources-generic-schedule backup.
+const BackupLabel = "cluster.open-cluster-management.io/backup"
+
 // Reason constants for the CollectorConfig Applied condition.
 const (
 	// CollectorConfigReasonApplied means all rules were processed and applied successfully.
--- a/api/v1alpha1/collectorconfig_types.go
+++ b/api/v1alpha1/collectorconfig_types.go
@@ -40,10 +40,6 @@ const IntegrationTeamLabel = "search.open-cluster-management.io/config-type"
 // IntegrationTeamLabelValue is the expected value for IntegrationTeamLabel.
 const IntegrationTeamLabelValue = "integration"
 
-// BackupLabel is the label applied to CollectorConfig CRs so they are included in ACM hub backups.
-// The search.open-cluster-management.io API group is excluded from the automatic resources backup,
-// but resources carrying this label are picked up by the acm-resources-generic-schedule backup.
-const BackupLabel = "cluster.open-cluster-management.io/backup"
 
 // Reason constants for the CollectorConfig Applied condition.
 const (
--- a/api/v1alpha1/collectorconfig_types.go
+++ b/api/v1alpha1/collectorconfig_types.go
@@ -40,7 +40,6 @@ const IntegrationTeamLabel = "search.open-cluster-management.io/config-type"
 // IntegrationTeamLabelValue is the expected value for IntegrationTeamLabel.
 const IntegrationTeamLabelValue = "integration"
 
-
 // Reason constants for the CollectorConfig Applied condition.
 const (
 	// CollectorConfigReasonApplied means all rules were processed and applied successfully.
--- a/controllers/create_collectorconfig.go
+++ b/controllers/create_collectorconfig.go
@@ -16,48 +16,27 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
-// ensureCollectorConfigsBackupLabel ensures that all source CollectorConfig CRs in the
-// operator namespace carry the ACM backup label so they survive a hub backup/restore cycle.
-//
+// addBackupLabel patches the backup label onto a CollectorConfig if it is missing.
 // The search.open-cluster-management.io API group is excluded from the automatic resources
 // backup (backup.go excludedAPIGroups). Resources labeled with backupLabel are picked up by
-// the acm-resources-generic-schedule backup instead.
-//
-// The only CollectorConfig excluded is merged-collector-config: it is operator-managed and
-// fully derived from the source configs on every reconcile, so backing it up would restore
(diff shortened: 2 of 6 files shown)


api/v1alpha1/collectorconfig_types.go
```diff
@@ -40,6 +40,11 @@ const IntegrationTeamLabel = "search.open-cluster-management.io/config-type"
 // IntegrationTeamLabelValue is the expected value for IntegrationTeamLabel.
 const IntegrationTeamLabelValue = "integration"
 
+// BackupLabel is the label applied to CollectorConfig CRs so they are included in ACM hub backups.
+// The search.open-cluster-management.io API group is excluded from the automatic resources backup,
+// but resources carrying this label are picked up by the acm-resources-generic-schedule backup.
+const BackupLabel = "cluster.open-cluster-management.io/backup"
```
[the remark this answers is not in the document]
sahare: Ran make generate and make manifests — both produce zero diffs. The change in collectorconfig_types.go is a string constant only; it does not add, remove, or modify any struct fields, kubebuilder markers, or RBAC annotations, so no regenerated artifacts are affected.

controllers/create_collectorconfig.go
```diff
@@ -16,9 +16,56 @@ import (
...
+
 const (
 	userCollectorConfigName   = "user-collector-config"
 	mergedCollectorConfigName = "merged-collector-config"
+
+	// backupLabel is the ACM backup label that causes a resource to be included
+	// in the acm-resources-generic-schedule backup. The search.open-cluster-management.io
+	// API group is excluded from automatic backups, so CollectorConfig CRs need this
+	// label to survive a hub backup/restore cycle.
+	backupLabel = "cluster.open-cluster-management.io/backup"
```
smcavey: Thanks for adding this as a const, could you replace the 2 hard-coded uses of this label in controllers/common and controllers/create_pgconfigmap with this? We have a large list of consts at the top of common.go, maybe this should go there since it's now being used in a few different files in the controllers dir

controllers/create_collectorconfig.go
```diff
@@ -16,9 +16,56 @@ import (
...
+	ctx context.Context,
+	namespace string,
+) (*reconcile.Result, error) {
+	ccList := &searchv1alpha1.CollectorConfigList{}
+	if err := r.List(ctx, ccList, client.InNamespace(namespace)); err != nil {
+		log.Error(err, "Could not list CollectorConfigs for backup label check")
+		return &reconcile.Result{}, err
+	}
+
+	for i := range ccList.Items {
```
smcavey: I'm not sure we want to backup each and every collectorconfig in the namespace. Presently merged-collector-config is created from the explicitly named "user-collector-config" plus all integration team labeled "search.open-cluster-management.io/config-type: integratation" collector-configs. _But_ we also might want to back them up all when we support override configs, users might have a custom config they want on a specific set of clusters....

smcavey: As discussed, current implementation is fine 😄 resolving. This captures any future configs including override ones.

controllers/create_collectorconfig.go
```diff
@@ -16,6 +16,50 @@ import (
...
+// fully derived from the source configs on every reconcile, so backing it up would restore
+// a stale snapshot rather than the sources. All other CollectorConfigs — user-collector-config,
+// integration team configs, and any future per-cluster override configs — are source configs
+// that users expect to survive a backup/restore and should always be labeled.
+func (r *SearchReconciler) ensureCollectorConfigsBackupLabel(
+	ctx context.Context,
+	namespace string,
+) (*reconcile.Result, error) {
+	ccList := &searchv1alpha1.CollectorConfigList{}
+	if err := r.List(ctx, ccList, client.InNamespace(namespace)); err != nil {
```
jlpadilla: The `createOrUpdateMergedCollectorConfig()` function below is also calling the API to list. Could we avoid the additional API request here?

jlpadilla (approved):
/lgtm

PR #740, opened 2026-06-18
ACM-35522: Webhook validation and merge protection for exclude action

sahare:
## Summary

Webhook validation and merge protection for `action: exclude` in CollectorConfig (ACM-35522).

**Companion PR:** stolostron/search-collector#888 (collector-side enforcement)

### `api/v1alpha1/collectorconfig_webhook.go`

**1. Wildcard bypass fix in `validateExcludeRule`:**
`kinds: ["*"]` on a protected apiGroup would effectively exclude ManagedCluster/Namespace just as reliably as naming them directly. Now rejected:
- `exclude kinds:["*"] apiGroups:["cluster.open-cluster-management.io"]` → rejected
- `exclude kinds:["*"] apiGroups:["*"]` → rejected (global wildcard)
- `exclude kinds:["*"] apiGroups:["apps"]` → accepted (no protected resources in apps)

**2. Integration config overlap check at admission (`validateExcludeAgainstIntegrationConfigs`):**
The webhook now lists CollectorConfigs labeled `search.open-cluster-management.io/config-type: integration` and rejects user exclude rules that conflict with any integration team include — surfacing the error at `kubectl apply` time rather than silently dropping at operator merge time.

This replaces the idea of expanding `protectedKinds` with a static list of resources (Argo, Kyverno, CNV, Policy, etc.), which would be fragile. Protection grows automatically as integration teams ship their own labeled CollectorConfigs. The check is skipped when `webhookClient` is nil (unit tests) or when no integration configs exist.

### `controllers/create_collectorconfig.go`

**Error message wording:** Status condition message when a user exclude is dropped: `"integration team has include for the same resource"` → `"resource is necessary for system functionality"`.

## Protection layers

```
User applies exclude on user-collector-config
        ↓
Webhook (admission)
  • Reject ManagedCluster/Namespace by name
  • Reject wildcard kinds on protected apiGroups
  • Reject if overlaps any integration team include (dynamic check)
        ↓
Operator merge
  • Drop overlapping user excludes (already done — second layer)
  • Write Applied=False / RulesSkipped status condition
        ↓
search-collector
  • IsResourceExcluded at informer layer (last-match-wins)
```

## Test plan

- `go test ./api/... ./controllers/...` — all tests pass, 7 new webhook tests
- `golangci-lint` — 0 issues
- E2E on cluster pending (see Jira ACM-35522 for full test matrix)

### DIFF
--- a/api/v1alpha1/collectorconfig_webhook.go
+++ b/api/v1alpha1/collectorconfig_webhook.go
@@ -259,38 +259,60 @@ func validateExcludeRule(rule *CollectionRule, path *field.Path) field.ErrorList
 	// Reject exclusion of protected resource types.
 	// Check both specific kind names and wildcard kinds — apiGroups:["cluster.open-cluster-management.io"]
 	// kinds:["*"] would exclude ManagedCluster just as effectively as naming it explicitly.
+	allErrs = append(allErrs, validateProtectedKinds(rule, path)...)
+	return allErrs
+}
+
+// validateProtectedKinds checks that an exclude rule does not target ManagedCluster or Namespace,
+// either by name or via a wildcard kind on their apiGroup.
+func validateProtectedKinds(rule *CollectionRule, path *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
 	for _, kind := range rule.ResourceSelector.Kinds {
 		if kind == "*" {
-			// Wildcard kind: reject if any targeted apiGroup contains a protected kind.
-			for protectedKind, protectedGroup := range protectedKinds {
-				for _, apiGroup := range rule.ResourceSelector.APIGroups {
-					if apiGroup == "*" || apiGroup == protectedGroup {
-						allErrs = append(allErrs, field.Invalid(
-							path.Child("resourceSelector", "apiGroups"),
-							apiGroup,
-							"cannot exclude all kinds in this apiGroup — it contains "+
-								protectedKind+", which search depends on for RBAC and cluster-scoped queries",
-						))
-						break
-					}
-				}
-			}
+			allErrs = append(allErrs, validateWildcardKindAgainstProtected(rule.ResourceSelector.APIGroups, path)...)
 			continue
 		}
 		if protectedGroup, protected := protectedKinds[kind]; protected {
-			for _, apiGroup := range rule.ResourceSelector.APIGroups {
-				if apiGroup == protectedGroup || apiGroup == "*" {
-					allErrs = append(allErrs, field.Invalid(
-						path.Child("resourceSelector", "kinds"),
-						kind,
-						"cannot exclude "+kind+" — search depends on it for RBAC and cluster-scoped queries",
-					))
-					break
-				}
+			allErrs = append(allErrs, validateSpecificKindAgainstProtected(kind, protectedGroup, rule.ResourceSelector.APIGroups, path)...)
+		}
+	}
+	return allErrs
+}
+
+// validateWildcardKindAgainstProtected rejects kinds:["*"] when the rule's apiGroups contain
+// a group that holds a protected kind (e.g. cluster.open-cluster-management.io → ManagedCluster).
+func validateWildcardKindAgainstProtected(apiGroups []string, path *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+	for protectedKind, protectedGroup := range protectedKinds {
+		for _, apiGroup := range apiGroups {
+			if apiGroup == "*" || apiGroup == protectedGroup {
+				allErrs = append(allErrs, field.Invalid(
+					path.Child("resourceSelector", "apiGroups"),
+					apiGroup,
+					"cannot exclude all kinds in this apiGroup — it contains "+
+						protectedKind+", which search depends on for RBAC and cluster-scoped queries",
+				))
+				break
 			}
 		}
 	}
+	return allErrs
+}
 
+// validateSpecificKindAgainstProtected rejects a specific protected kind when the rule's
+// apiGroups match the kind's group (including the global wildcard "*").
+func validateSpecificKindAgainstProtected(kind, protectedGroup string, apiGroups []string, path *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+	for _, apiGroup := range apiGroups {
+		if apiGroup == protectedGroup || apiGroup == "*" {
+			allErrs = append(allErrs, field.Invalid(
+				path.Child("resourceSelector", "kinds"),
+				kind,
+				"cannot exclude "+kind+" — search depends on it for RBAC and cluster-scoped queries",
+			))
+			break
+		}
+	}
 	return allErrs
 }
 
--- a/api/v1alpha1/collectorconfig_webhook.go
+++ b/api/v1alpha1/collectorconfig_webhook.go
@@ -273,7 +273,8 @@ func validateProtectedKinds(rule *CollectionRule, path *field.Path) field.ErrorL
 			continue
 		}
 		if protectedGroup, protected := protectedKinds[kind]; protected {
-			allErrs = append(allErrs, validateSpecificKindAgainstProtected(kind, protectedGroup, rule.ResourceSelector.APIGroups, path)...)
+			errs := validateSpecificKindAgainstProtected(kind, protectedGroup, rule.ResourceSelector.APIGroups, path)
+			allErrs = append(allErrs, errs...)
 		}
 	}
 	return allErrs
@@ -301,7 +302,9 @@ func validateWildcardKindAgainstProtected(apiGroups []string, path *field.Path) 
 
 // validateSpecificKindAgainstProtected rejects a specific protected kind when the rule's
 // apiGroups match the kind's group (including the global wildcard "*").
-func validateSpecificKindAgainstProtected(kind, protectedGroup string, apiGroups []string, path *field.Path) field.ErrorList {
+func validateSpecificKindAgainstProtected(
+	kind, protectedGroup string, apiGroups []string, path *field.Path,
+) field.ErrorList {
 	var allErrs field.ErrorList
 	for _, apiGroup := range apiGroups {
 		if apiGroup == protectedGroup || apiGroup == "*" {
--- a/api/v1alpha1/collectorconfig_webhook.go
+++ b/api/v1alpha1/collectorconfig_webhook.go
@@ -12,6 +12,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
@@ -20,6 +21,11 @@ import (
 // log is for logging in this package.
 var collectorconfiglog = logf.Log.WithName("collectorconfig-resource")
 
+// webhookClient is the Kubernetes client used to list integration team CollectorConfigs
+// during admission. Set once in SetupWebhookWithManager; nil in unit tests that don't
+// register a manager (the integration-config check is skipped when nil).
+var webhookClient client.Client
+
 // Precompiled validation patterns.
 var (
 	fieldNamePattern   = regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9\-_.]*$`)
@@ -27,6 +33,7 @@ var (
 )
 
 func (r *CollectorConfig) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	webhookClient = mgr.GetClient()
 	return ctrl.NewWebhookManagedBy(mgr).
 		For(r).
 		WithValidator(r).
@@ -50,8 +57,11 @@ func (r *CollectorConfig) ValidateCreate(ctx context.Context, obj runtime.Object
 		return nil, err
 	}
 
-	err := cc.validateCollectorConfig()
-	return nil, err
+	if err := cc.validateCollectorConfig(); err != nil {
+		return nil, err
+	}
+
+	return nil, validateExcludeAgainstIntegrationConfigs(ctx, cc)
 }
 
 // ValidateUpdate implements webhook.CustomValidator so a webhook will be registered for the type
@@ -74,8 +84,11 @@ func (r *CollectorConfig) ValidateUpdate(ctx context.Context, oldObj, newObj run
 		return nil, err
 	}
 
-	err := cc.validateCollectorConfig()
-	return nil, err
+	if err := cc.validateCollectorConfig(); err != nil {
+		return nil, err
+	}
+
+	return nil, validateExcludeAgainstIntegrationConfigs(ctx, cc)
 }
 
 // ValidateDelete implements webhook.CustomValidator so a webhook will be registered for the type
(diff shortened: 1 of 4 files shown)


controllers/create_collectorconfig_test.go
sahare, line 1: **Merge conflict — PR needs rebase before this can land.**

There are 6 unresolved conflict markers in this file (starting at line 450), left over from merging with the `TestBackupLabel_*` tests added to `main`. Running `go test ./controllers/...` fails immediately with `expected declaration, found '<<'`.

Please rebase onto the latest `main` and resolve the conflicts, keeping both the backup-label tests (from main) and the new `TestMerge_UserExclude*` tests (from this PR).

api/v1alpha1/collectorconfig_webhook.go
```diff
@@ -177,6 +182,64 @@ func (r *CollectorConfig) validateCollectorConfig() error {
...
+
+	// Reject fields, collectConditions, and fieldSuffix on exclude rules.
+	if len(rule.Fields) > 0 {
+		allErrs = append(allErrs, field.Invalid(
+			path.Child("fields"),
+			rule.Fields,
+			"fields cannot be specified on an exclude rule",
+		))
+	}
+	if rule.CollectConditions != nil && *rule.CollectConditions {
```
sahare: **Nil-check is too narrow — `collectConditions: false` slips through.**

The current guard:
```go
if rule.CollectConditions != nil && *rule.CollectConditions {
```
only rejects `collectConditions: true`. A user who writes `collectConditions: false` on an exclude rule has made a configuration mistake (the field is meaningless on an exclude), but their CR is accepted without error.

Consider changing to:
```suggestion
	if rule.CollectConditions != nil {
```
Any explicit pointer value — `true` or `false` — on an exclude rule signals a misunderstanding and should be rejected. The natural default is `nil` (unset), which already passes.

controllers/create_collectorconfig.go
```diff
@@ -78,6 +78,51 @@ func (r *SearchReconciler) ensureWebhookCAInjection(ctx context.Context) error {
...
+	return setsIntersect(a.Kinds, b.Kinds)
+}
+
+// setsIntersect returns true when two string slices share at least one element,
+// treating "*" as a universal match for all values on the other side.
+func setsIntersect(a, b []string) bool {
+	for _, x := range a {
+		for _, y := range b {
+			if x == "*" || y == "*" || x == y {
+				return true
```
sahare, line 122: **Silent drop is invisible to the user — consider surfacing it via status conditions.**

`log.V(2)` is debug-level; operators running at the default verbosity will never see this message. The user who wrote `user-collector-config` has no way to discover that their exclude rule was silently dropped.

The `CollectorConfigStatus.Conditions` type already has a `CollectorConfigReasonRulesSkipped` constant for exactly this scenario. I would suggest updating the `Applied` condition on `userCC` to `False` / `RulesSkipped` with a message listing the dropped rules — something like:

```go
// After the merge loop, if any rules were dropped, patch userCC status.
```

If a full status patch is out of scope for this PR, at minimum please elevate this to `log.Info` (V(0)) so it shows up in normal production operator logs without requiring `--zap-log-level=2`.

api/v1alpha1/collectorconfig_webhook.go
```diff
@@ -168,6 +168,11 @@ func (r *CollectorConfig) validateCollectorConfig() error {
 				))
 			}
 		}
+
+		// Validate exclude-specific constraints
+		if rule.Action == ActionExclude {
+			allErrs = append(allErrs, validateExcludeRule(&rule, rulePath)...)
```
sahare, line 200: **Heads up: calling `validateExcludeRule` after the fieldSuffix/wildcard checks produces double errors in some cases.**

For `action: exclude, fieldSuffix: Bad` the validator emits two errors:
1. Pre-existing fieldSuffix format check (line 162–169): "must contain only lowercase..."
2. Here via `validateExcludeRule` line 213: "fieldSuffix cannot be specified on an exclude rule"

The second message is the correct/helpful one. Consider guarding the format check with `rule.Action != ActionExclude` so only the meaningful error fires.

Same double-error applies to `action: exclude, kinds: ["*"], fields: [...]`: the wildcard+fields check (line 122) fires before the fields check inside `validateExcludeRule`.

api/v1alpha1/collectorconfig_webhook_test.go
sahare, line 1: **Missing test coverage: wildcard apiGroup with protected kinds.**

The existing tests cover `APIGroups: ["cluster.open-cluster-management.io"], Kinds: ["ManagedCluster"]` but there is no test for `APIGroups: ["*"], Kinds: ["ManagedCluster"]` or `APIGroups: ["*"], Kinds: ["Namespace"]`. The code at line 228 handles this correctly (`apiGroup == "*"` check), but a regression test would catch any future refactor.

Suggested additions:
```go
// Reject exclude targeting ManagedCluster via wildcard apiGroup.
func TestRejectExcludeManagedClusterWildcardGroup(t *testing.T) {
    c := validConfig()
    c.Spec.CollectionRules[0] = CollectionRule{
        Action: ActionExclude,
        ResourceSelector: ResourceSelector{
            APIGroups: []string{"*"},
            Kinds:     []string{"ManagedCluster"},
        },
    }
    _, err := c.ValidateCreate(context.Background(), c)
    assert.Error(t, err)
    assert.Contains(t, err.Error(), "cannot exclude ManagedCluster")
}
```
(Same for `Namespace`.)

api/v1alpha1/collectorconfig_webhook.go
```diff
@@ -177,6 +182,64 @@ func (r *CollectorConfig) validateCollectorConfig() error {
 	return allErrs.ToAggregate()
 }
 
+// protectedKinds lists resource types the search RBAC engine depends on.
+// Excluding them would break per-cluster and namespace-scoped access control.
+var protectedKinds = map[string]string{
+	"ManagedCluster": "cluster.open-cluster-management.io",
```
sahare, line 219: **Consider documenting why only these two resources are protected.**

The comment says the RBAC engine depends on ManagedCluster and Namespace, but does not explain what would break for a cluster admin who excludes them (e.g., "search-api uses ManagedCluster names as cluster identifiers in subjectAccessReviews; Namespace is used to scope namespace-scoped resource filtering").

Also worth confirming: does the RBAC engine also depend on `ManagedClusterSet` / `ManagedClusterSetBinding` (`cluster.open-cluster-management.io`)? If cluster-set-scoped access control is in scope, those may need protection too. If they do not need to be in this list, a brief explanation of why helps future maintainers avoid second-guessing.

sahare (commented):
Good overall direction and the core logic is sound — `setsIntersect` handles wildcards bidirectionally correctly, `validateExcludeRule` covers the right semantic constraints, and the merge-protection design (operator-time drop rather than collector-time) is the right call.

Two things need to be addressed before merge:
1. **Rebase required** — unresolved conflict markers in `create_collectorconfig_test.go` prevent compilation.
2. **Silent drop needs user visibility** — the `log.V(2)` on the skipped exclude rule is invisible in production; at minimum bump to `log.Info`, ideally update the `user-collector-config` Applied condition with `RulesSkipped`.

See inline comments for the rest.

sahare (commented):
**Second-pass review — all first-pass issues are correctly addressed. Ready for LGTM/approve.**

### What was verified

- **Rebase conflict resolved** — `TestBackupLabel_*` (from main) and `TestMerge_UserExclude*` (from this PR) coexist cleanly; `go test ./api/... ./controllers/...` passes.
- **`collectConditions` nil-check fixed** — `webhook.go:215` now rejects any explicitly set pointer value (`true` or `false`), not just `true`.
- **Double-error fix** — wildcard+fields and fieldSuffix format checks are now guarded by `rule.Action == ActionInclude`. An exclude rule with an invalid/non-empty fieldSuffix is still rejected — the format check is simply skipped and `validateExcludeRule` fires the more accurate single error. No silent acceptance gap introduced.
- **Silent drop elevated to `log.Info`** — visible in production logs without `--zap-log-level=2`.
- **`protectedKinds` documented** — comment now explains why ManagedCluster and Namespace are protected, and explicitly states why ManagedClusterSet is not.
- **Two new tests added** — `TestRejectExcludeManagedClusterViaWildcardAPIGroup` and `TestRejectExcludeWithCollectConditionsFalse`.

### Minor remaining gaps (non-blocking)

1. No test for `APIGroups: ["*"], Kinds: ["Namespace"]` — Namespace wildcard exercises the same code path as the ManagedCluster wildcard test already present. Defensible to defer.
2. No merge test for mixed exclude rules (one overlapping, one not) — code is correct; a regression test would add safety. Also defensible to defer.

Neither is a correctness issue.

api/v1alpha1/collectorconfig_webhook.go
```diff
@@ -177,6 +182,64 @@ func (r *CollectorConfig) validateCollectorConfig() error {
...
+			rule.Fields,
+			"fields cannot be specified on an exclude rule",
+		))
+	}
+	if rule.CollectConditions != nil && *rule.CollectConditions {
+		allErrs = append(allErrs, field.Invalid(
+			path.Child("collectConditions"),
+			*rule.CollectConditions,
+			"collectConditions cannot be set on an exclude rule",
+		))
```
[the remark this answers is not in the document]
sahare: Already fixed in commit d375692: the condition is now `if rule.CollectConditions != nil` (checking presence only, not value).

controllers/create_collectorconfig.go
sahare: Implemented the full fix in commit 42aad3e: `updateUserCCStatus` now patches the Applied condition on `user-collector-config` with `Applied=False / RulesSkipped` and a message listing the dropped rules when any exclude was blocked by integration team protection. When no rules are dropped, it sets `Applied=True`. The status subresource is updated via `r.Status().Update()`, consistent with how the collector writes conditions. A new test `TestMerge_UserExcludeDropped_StatusConditionSet` covers this path.

controllers/create_collectorconfig_test.go
sahare: Fixed in rebase — all 6 conflict markers resolved in the merge, backup-label tests (TestBackupLabel_*) and new exclude tests (TestMerge_UserExclude*) coexist cleanly. go test ./controllers/... passes.

api/v1alpha1/collectorconfig_webhook.go
sahare: Fixed in commit d375692 — both the fieldSuffix format check and the wildcard+fields check are now wrapped in if rule.Action == ActionInclude so exclude rules get a single precise error from validateExcludeRule only.

api/v1alpha1/collectorconfig_webhook_test.go
sahare: Added TestRejectExcludeManagedClusterViaWildcardAPIGroup and TestRejectExcludeWithCollectConditionsFalse in the fix commit. The wildcard apiGroup test is now in place.

api/v1alpha1/collectorconfig_webhook.go
sahare: Added in commit d375692 — the comment now explicitly explains why only ManagedCluster and Namespace are listed, and why ManagedClusterSet/ManagedClusterSetBinding are NOT (the RBAC engine uses placement/policy, not search directly).

sahare:
@coderabbitai review

smcavey:
Just FYI regarding stale records after configuration changes: https://redhat.atlassian.net/browse/ACM-20047 is going to handle that

controllers/create_collectorconfig.go
```diff
@@ -126,13 +213,33 @@ func (r *SearchReconciler) createOrUpdateMergedCollectorConfig(
...
 	if err == nil {
-		mergedSpec.CollectionRules = append(mergedSpec.CollectionRules, userCC.Spec.CollectionRules...)
+		var droppedRuleMessages []string
+		for _, rule := range userCC.Spec.CollectionRules {
+			// Integration team wins: drop user exclude rules that overlap with any
+			// integration team include so they cannot suppress integration-required collection.
+			if rule.Action == searchv1alpha1.ActionExclude &&
+				excludeOverlapsIntegrationIncludes(rule, teamConfigs.Items) {
+				msg := fmt.Sprintf(
+					"exclude rule for kinds %v (apiGroups %v) was dropped — integration team has include for the same resource",
```
smcavey: Do we think users will know what "integration team" means? Maybe we generalize to something like "unable to honor exclude rule for kinds %v ( apiGroups %v) - resource necessary for system functionality" ?

api/v1alpha1/collectorconfig_webhook.go
```diff
@@ -186,6 +195,76 @@ func (r *CollectorConfig) validateCollectorConfig() error {
...
 }
 
+// protectedKinds lists resource types the search RBAC engine depends on.
+// Excluding them would break per-cluster and namespace-scoped access control:
+//   - ManagedCluster: used to scope search results to clusters a user can access
+//   - Namespace: used to scope search results to namespaces a user can access
+//
+// ManagedClusterSet and ManagedClusterSetBinding are NOT listed here because the
+// RBAC engine does not query them directly — they are used by placement/policy, not search.
+var protectedKinds = map[string]string{
```
smcavey, line 218: We need to consider how to handle the resources already hard-coded into the search-collector that are necessary for both Search to function and integration teams. i.e. https://github.com/stolostron/search-collector/tree/main/pkg/transforms contains things like argoapplication.go, kyverno.go, policy.go, and genericResourceConfig.go contain lots of fields necessary for Search and integrations to function, not sure there's a good way to handle it but I don't think for 5.0 we should expect Policy, CNV, etc to provide us with the historical list of includes that must be protected

smcavey: I think we need to expand this list unless there's a better way

jlpadilla: Agree. A hardcoded list will work for now, but most likely we'll need a better approach soon.

jlpadilla: Actually, I see `func excludeOverlapsIntegrationIncludes()`.
That's a very good start. The only problem is that these integration configs aren't defined yet, so we need to fill the gaps until we get there.

api/v1alpha1/collectorconfig_webhook.go
```diff
@@ -328,3 +461,88 @@ func rejectIfProtected(ctx context.Context, cc *CollectorConfig) error {
...
+}
+
+// webhookSetsIntersect returns true when two string slices share at least one element,
+// treating "*" as a universal match for all values on the other side.
+// This mirrors controllers.setsIntersect but is local to the webhook package to
+// avoid an import cycle.
+func webhookSetsIntersect(a, b []string) bool {
+	for _, x := range a {
+		for _, y := range b {
+			if x == "*" || y == "*" || x == y {
```
smcavey, line 621: I don't see any tests that hit the wildcard "*" kind path. We can add:
```
// Reject user wildcard kinds exclude that overlaps an integration team include.
func TestRejectWildcardKindsExcludeOverlapsIntegrationInclude(t *testing.T) {
	buildFakeWebhookClient(t,
		integrationCC("grc-config", "default", "policy.open-cluster-management.io", []string{"Policy"}),
	)
	c := validConfig()
	c.Namespace = "default"
	c.Spec.CollectionRules[0] = CollectionRule{
		Action: ActionExclude,
		ResourceSelector: ResourceSelector{
			APIGroups: []string{"policy.open-cluster-management.io"},
			Kinds:     []string{"*"},
		},
	}
	_, err := c.ValidateCreate(context.Background(), c)
	assert.Error(t, err, "exclude wildcard kinds overlapping integration include must be rejected")
	assert.Contains(t, err.Error(), "necessary for system functionality")
}
```

api/v1alpha1/collectorconfig_webhook.go
```diff
@@ -186,6 +208,117 @@ func (r *CollectorConfig) validateCollectorConfig() error {
...
+			"fields cannot be specified on an exclude rule",
+		))
+	}
+	if rule.CollectConditions != nil {
+		allErrs = append(allErrs, field.Invalid(
+			path.Child("collectConditions"),
+			*rule.CollectConditions,
+			"collectConditions cannot be set on an exclude rule",
+		))
+	}
```
smcavey, line 295: If this PR isn't already rebased on main we're going to be missing a smiilar check against rule.CollectAdditionalPrinterColumnsPriority

api/v1alpha1/collectorconfig_webhook.go
smcavey: https://github.com/stolostron/search-v2-operator/pull/740/changes/8847d113d4722766ad9ea6402d544f5e6277144d#diff-ba5ef31a37d93a0dfe0181d0659002e1196b57fbd7b6c1327be71771f73ea5d7R231-R274 should cover the gaps until we come back and start migrating integration team requirements into integration collectorconfigs

smcavey: https://redhat.atlassian.net/browse/ACM-37052 regarding integration team collectorconfigs and removing the stopgap of apiGroups list

PR #748, opened 2026-07-09
ACM-33145 - openapiv3 data type alignment

smcavey:
### Related Issue

https://issues.redhat.com/browse/ACM-33145

### Description of changes
- Removed number, slice, and mapString from collectorconfig supported datatypes.
- Added boolean, integer, and float to collectorconfig supported datatypes

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated supported field types for collection rules to a clearer set of values.
  * Improved validation so only the newly supported types are accepted.
  * Existing configurations using removed type values will no longer validate.
* **Tests**
  * Adjusted validation tests to cover the updated allowed and disallowed field types.
* **Chores**
  * Synced schema and published manifests with the new type list.

### DIFF
--- a/api/v1alpha1/collectorconfig_types.go
+++ b/api/v1alpha1/collectorconfig_types.go
@@ -9,15 +9,15 @@ import (
 // Important: Run the "make manifests" command to regenerate the manifests after you modify this file.
 
 // DataType represents the data type of a collected field
-// +kubebuilder:validation:Enum=bytes;slice;string;number;mapString
+// +kubebuilder:validation:Enum=bytes;string;integer;float;boolean
 type DataType string
 
 const (
-	DataTypeBytes     DataType = "bytes"
-	DataTypeSlice     DataType = "slice"
-	DataTypeString    DataType = "string"
-	DataTypeNumber    DataType = "number"
-	DataTypeMapString DataType = "mapString"
+	DataTypeBytes   DataType = "bytes"
+	DataTypeString  DataType = "string"
+	DataTypeInteger DataType = "integer"
+	DataTypeFloat   DataType = "float"
+	DataTypeBoolean DataType = "boolean"
 )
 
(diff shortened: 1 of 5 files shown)


api/v1alpha1/collectorconfig_types.go
```diff
@@ -9,15 +9,15 @@ import (
...
-// +kubebuilder:validation:Enum=bytes;slice;string;number;mapString
+// +kubebuilder:validation:Enum=bytes;string;integer;float;boolean
 type DataType string
 
 const (
-	DataTypeBytes     DataType = "bytes"
-	DataTypeSlice     DataType = "slice"
-	DataTypeString    DataType = "string"
-	DataTypeNumber    DataType = "number"
-	DataTypeMapString DataType = "mapString"
```
jlpadilla, line 20: "slice" and "mapString" are removed. Was this intentional?

smcavey: Yes it was. They're quite situational. I could consider leaving `slice` and renaming to `array` for alignment, but I don't think we need to expose "mapString" at minimum. DataTypeSlice is presently used on 2 resources, both of which for CNV: virtualmachine.kubevirt.io and virtualmachineinstance.kubevirt.io

smcavey: Even then, supporting an array may lead to wanting an arrays of integers, floats, strings, ... where internally it's only used for strings. Thinking integer, float, byte, string, and boolean should cover the vast amount of configuration requests if our present internal config is any indication

jlpadilla: ok, thanks for the explanation. It's good to keep it simple, if it's really needed we can add it later.

jlpadilla (approved):
/lgtm

PR #757, opened 2026-07-16
ACM-37052: Built-in integration CollectorConfigs, seeded at startup

sahare:
## ACM-37052: Built-in integration CollectorConfigs, seeded at startup

Integration teams (CNV, OLM, GRC, Kyverno, Gatekeeper, Argo, ACM app lifecycle) can now ship a
default `CollectorConfig` by dropping a plain YAML file into
`config/integration_collector_configs/` — no Go code required. This is the follow-through on the
`protectedAPIGroups` safety net from the exclude-action work (ACM-35522), which was explicitly
temporary "until integration teams ship labeled CollectorConfig CRs." That static list shrinks
from 28 groups down to 6 now that most of it is covered by these real configs.

**Approach:** the operator embeds these YAML files at build time and, once per process at manager
startup (not on every reconcile), creates or unconditionally overwrites each one on the cluster.
Because this only happens at startup, a team can freely edit their canonical config and it
persists for the life of that pod — the next restart/upgrade resets it to whatever's currently
shipped. A team that wants a change to survive restarts before it ships officially just creates a
differently-named CollectorConfig instead (still picked up by the existing label-based merge
step) — the seeder only touches its fixed set of known names.

Full design writeup in `docs/ARCHITECTURE.md`.

## Summary by CodeRabbit

* **New Features**
  * Added built-in integration `CollectorConfig` manifests for app lifecycle, Argo CD, CNV, OLM, Gatekeeper, GRC, and Kyverno.
  * Operator now seeds these configurations at startup with retry until webhook readiness; canonical configs are kept in sync while alternate overrides are preserved.
* **Bug Fixes**
  * Updated webhook validation so integration-owned exclusions are handled correctly, including for non-service-account users.
* **Documentation**
  * Expanded architecture docs for startup seeding, namespace discovery, and pause behavior.
* **Chores**
  * Build now packages the embedded integration configuration assets; lint timeout increased.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -204,7 +204,7 @@ catalog-push: ## Push a catalog image.
 lint:
 	GOPATH=$(go env GOPATH)
 	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b "${GOPATH}/bin" v2.9.0
-	CGO_ENABLED=1 GOGC=25 golangci-lint run --timeout=3m
+	CGO_ENABLED=1 GOGC=25 golangci-lint run --timeout=8m
 	gosec ./...
 
 # Setup local development environment.
--- a/Makefile
+++ b/Makefile
@@ -204,7 +204,7 @@ catalog-push: ## Push a catalog image.
 lint:
 	GOPATH=$(go env GOPATH)
 	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b "${GOPATH}/bin" v2.9.0
-	CGO_ENABLED=1 GOGC=25 golangci-lint run --timeout=8m
+	CGO_ENABLED=1 GOGC=25 golangci-lint run --timeout=3m
 	gosec ./...
 
 # Setup local development environment.
--- a/config/integration_collector_configs/app-lifecycle.yaml
+++ b/config/integration_collector_configs/app-lifecycle.yaml
@@ -1,0 +1,22 @@
+# Integration CollectorConfig for ACM Application Lifecycle (multicloud-operators subscription
+# stack: Channel, Deployable, HelmRelease, PlacementBinding, Subscription, and the upstream
+# sig-application "Application" kind). Populated from resources already hard-coded in
+# search-collector's typed transforms (channel.go, appdeployable.go, helmrelease.go,
+# placementbinding.go, subscription.go, application.go) before this CollectorConfig existed.
+# See ACM-37052.
+apiVersion: search.open-cluster-management.io/v1alpha1
+kind: CollectorConfig
+metadata:
+  name: app-lifecycle-integration-collector-config
+  namespace: open-cluster-management
+  labels:
+    search.open-cluster-management.io/config-type: integration
+spec:
+  collectionRules:
+    - action: include
+      resourceSelector:
+        apiGroups:
+          - apps.open-cluster-management.io
+          - app.k8s.io
+        kinds:
+          - "*"
--- a/config/integration_collector_configs/app-lifecycle.yaml
+++ b/config/integration_collector_configs/app-lifecycle.yaml
@@ -11,6 +11,7 @@ metadata:
   namespace: open-cluster-management
   labels:
     search.open-cluster-management.io/config-type: integration
+    cluster.open-cluster-management.io/backup: ""
 spec:
   collectionRules:
     - action: include
--- a/config/integration_collector_configs/app-lifecycle.yaml
+++ b/config/integration_collector_configs/app-lifecycle.yaml
@@ -4,6 +4,9 @@
 # sig-application "Application" kind). Populated from resources already hard-coded in
 # search-collector's typed transforms (channel.go, appdeployable.go, helmrelease.go,
 # placementbinding.go, subscription.go, application.go) before this CollectorConfig existed.
+#
+# Managed by the operator — reset to this default on restart unless annotated with
+# search.open-cluster-management.io/manual-override. See docs/ARCHITECTURE.md.
 apiVersion: search.open-cluster-management.io/v1alpha1
 kind: CollectorConfig
 metadata:
--- a/config/integration_collector_configs/app-lifecycle.yaml
+++ b/config/integration_collector_configs/app-lifecycle.yaml
@@ -1,14 +1,13 @@
+# Copyright Contributors to the Open Cluster Management project
 # Integration CollectorConfig for ACM Application Lifecycle (multicloud-operators subscription
 # stack: Channel, Deployable, HelmRelease, PlacementBinding, Subscription, and the upstream
 # sig-application "Application" kind). Populated from resources already hard-coded in
 # search-collector's typed transforms (channel.go, appdeployable.go, helmrelease.go,
 # placementbinding.go, subscription.go, application.go) before this CollectorConfig existed.
-# See ACM-37052.
 apiVersion: search.open-cluster-management.io/v1alpha1
 kind: CollectorConfig
 metadata:
-  name: app-lifecycle-integration-collector-config
-  namespace: open-cluster-management
+  name: app-lifecycle-integration
   labels:
     search.open-cluster-management.io/config-type: integration
     cluster.open-cluster-management.io/backup: ""
--- a/config/integration_collector_configs/cnv.yaml
+++ b/config/integration_collector_configs/cnv.yaml
@@ -15,6 +15,7 @@ metadata:
   namespace: open-cluster-management
   labels:
     search.open-cluster-management.io/config-type: integration
+    cluster.open-cluster-management.io/backup: ""
 spec:
   collectionRules:
     - action: include
--- a/config/integration_collector_configs/cnv.yaml
+++ b/config/integration_collector_configs/cnv.yaml
@@ -1,0 +1,34 @@
+# Integration CollectorConfig for CNV (Container Native Virtualization / KubeVirt) and the
+# storage resources it depends on. Populated from resources already hard-coded in
+# search-collector/pkg/transforms/genericResourceConfig.go and transformer.go before this
+# CollectorConfig existed. See ACM-37052.
+#
+# This config is created by the search-v2-operator if not already present on the cluster
+# (config-hash annotation tracks the shipped version — see docs/ARCHITECTURE.md). Once created,
+# CNV is free to edit it directly; the operator will only overwrite it again if a future
+# release ships an updated default AND this object hasn't been locally customized since.
+apiVersion: search.open-cluster-management.io/v1alpha1
+kind: CollectorConfig
+metadata:
+  name: cnv-integration-collector-config
+  namespace: open-cluster-management
+  labels:
+    search.open-cluster-management.io/config-type: integration
+spec:
+  collectionRules:
+    - action: include
+      resourceSelector:
+        apiGroups:
+          - kubevirt.io
+          - cdi.kubevirt.io
+          - migrations.kubevirt.io
+          - clone.kubevirt.io
+          - instancetype.kubevirt.io
+          - snapshot.kubevirt.io
+          - networkaddonsoperator.network.kubevirt.io
+          - k8s.cni.cncf.io
+          - storage.k8s.io
+          - snapshot.storage.k8s.io
+          - snapshot.storage.kubevirt.io
+        kinds:
+          - "*"
--- a/config/integration_collector_configs/cnv.yaml
+++ b/config/integration_collector_configs/cnv.yaml
@@ -3,10 +3,11 @@
 # search-collector/pkg/transforms/genericResourceConfig.go and transformer.go before this
 # CollectorConfig existed. See ACM-37052.
 #
-# This config is created by the search-v2-operator if not already present on the cluster
-# (config-hash annotation tracks the shipped version — see docs/ARCHITECTURE.md). Once created,
-# CNV is free to edit it directly; the operator will only overwrite it again if a future
-# release ships an updated default AND this object hasn't been locally customized since.
+# This config is created by the search-v2-operator once per process, at startup, and reset to
+# this shipped default on every restart/upgrade — see docs/ARCHITECTURE.md. CNV can edit it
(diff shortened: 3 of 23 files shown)


config/integration_collector_configs/cnv.yaml
```diff
@@ -0,0 +1,35 @@
...
+          - clone.kubevirt.io
+          - instancetype.kubevirt.io
+          - snapshot.kubevirt.io
+          - networkaddonsoperator.network.kubevirt.io
+          - k8s.cni.cncf.io
+          - storage.k8s.io
+          - snapshot.storage.k8s.io
+          - snapshot.storage.kubevirt.io
+        kinds:
+          - "*"
```
smcavey, line 33: I'm happy with this approach of clearing out the `protectedAPIGroups` map and moving them to these. In a follow up or eventually we can should make these more granular like https://github.com/stolostron/search-collector/blob/main/pkg/transforms/genericResourceConfig.go#L355-L384 using specific fields so that as these get adopted by the integration team they have a singular file to see all that's configured for their integration so far. But in a followup

controllers/create_integration_collectorconfigs.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// Sort for deterministic, readable logs (fs.ReadDir is already sorted, but be explicit).
+	sort.Slice(entries, func(i, j int) bool { return entries[i].Name() < entries[j].Name() })
+
+	for _, entry := range entries {
+		if entry.IsDir() {
+			continue
+		}
+		if err := applyOneIntegrationCollectorConfig(ctx, c, namespace, fsys, dir, entry.Name()); err != nil {
+			return err
```
smcavey: Why don't we log and continue on failed attempts to apply the collectorconfigs? A new change modifying integration config a.yaml could break every other integration

config/integration_collector_configs/app-lifecycle.yaml
```diff
@@ -0,0 +1,22 @@
...
+# sig-application "Application" kind). Populated from resources already hard-coded in
+# search-collector's typed transforms (channel.go, appdeployable.go, helmrelease.go,
+# placementbinding.go, subscription.go, application.go) before this CollectorConfig existed.
+# See ACM-37052.
+apiVersion: search.open-cluster-management.io/v1alpha1
+kind: CollectorConfig
+metadata:
+  name: app-lifecycle-integration-collector-config
+  namespace: open-cluster-management
+  labels:
```
smcavey, line 14: We should add the backup label to all of these

config/integration_collector_configs/app-lifecycle.yaml
```diff
@@ -0,0 +1,22 @@
+# Integration CollectorConfig for ACM Application Lifecycle (multicloud-operators subscription
+# stack: Channel, Deployable, HelmRelease, PlacementBinding, Subscription, and the upstream
+# sig-application "Application" kind). Populated from resources already hard-coded in
+# search-collector's typed transforms (channel.go, appdeployable.go, helmrelease.go,
+# placementbinding.go, subscription.go, application.go) before this CollectorConfig existed.
+# See ACM-37052.
+apiVersion: search.open-cluster-management.io/v1alpha1
+kind: CollectorConfig
+metadata:
```
smcavey, line 12: These also should have ownerReference to Search so they get cleaned up if/when Search gets torn down

config/integration_collector_configs/app-lifecycle.yaml
```diff
@@ -0,0 +1,23 @@
...
+  name: app-lifecycle-integration-collector-config
+  namespace: open-cluster-management
```
jlpadilla: Namespace must be variable. ACM could be deployed in a different namespace, QE uses `ocm` as their deploy namespace.

config/integration_collector_configs/app-lifecycle.yaml
```diff
@@ -0,0 +1,23 @@
+# Integration CollectorConfig for ACM Application Lifecycle (multicloud-operators subscription
+# stack: Channel, Deployable, HelmRelease, PlacementBinding, Subscription, and the upstream
+# sig-application "Application" kind). Populated from resources already hard-coded in
+# search-collector's typed transforms (channel.go, appdeployable.go, helmrelease.go,
+# placementbinding.go, subscription.go, application.go) before this CollectorConfig existed.
+# See ACM-37052.
+apiVersion: search.open-cluster-management.io/v1alpha1
+kind: CollectorConfig
+metadata:
+  name: app-lifecycle-integration-collector-config
```
jlpadilla: The suffix `-collector-config` in the name is redundant. I'd keep the names simple.

controllers/create_integration_collectorconfigs.go
```diff
@@ -0,0 +1,182 @@
...
+		return nil
+	}
+	found.Spec = desired.Spec
+	if found.Labels == nil {
+		found.Labels = map[string]string{}
+	}
+	// Always (re-)set the label, even when only the spec differed — a pre-existing config found
+	// without it would otherwise be invisible to the webhook's integration-overlap check and to
+	// the merge step's label-based discovery, silently letting conflicting user excludes through.
+	found.Labels[searchv1alpha1.IntegrationTeamLabel] = searchv1alpha1.IntegrationTeamLabelValue
```
jlpadilla, line 165: 🤖 I'm an AI agent

**Update path only enforces `IntegrationTeamLabel` — other labels from the embedded YAML (e.g. the backup label) are not restored if removed.**

On **Create** (line 114), all labels from `desired.Labels` are copied over, including `cluster.open-cluster-management.io/backup: ""`. On **Update** here, only `IntegrationTeamLabel` is re-applied. So if someone removes the backup label from a live integration CollectorConfig, the seeder will reset the spec and enforce the integration label, but will silently leave the backup label absent — it never comes back until the CR is deleted and re-created.

Suggested fix — merge all labels from the shipped YAML in the update path, then enforce the integration label on top:

```suggestion
	for k, v := range desired.Labels {
		found.Labels[k] = v
	}
	found.Labels[searchv1alpha1.IntegrationTeamLabel] = searchv1alpha1.IntegrationTeamLabelValue
```

This makes Create and Update symmetric: the shipped YAML's labels are always the source of truth.

config/integration_collector_configs/cnv.yaml
```diff
@@ -0,0 +1,36 @@
...
+# This config is created by the search-v2-operator once per process, at startup, and reset to
+# this shipped default on every restart/upgrade — see docs/ARCHITECTURE.md. CNV can edit it
+# directly and the edit will persist until the next restart, but to make a change survive
+# restarts before it's officially shipped here, create a differently-named CollectorConfig
+# (e.g. cnv-integration-collector-config-2) instead — the operator only manages this exact name.
+apiVersion: search.open-cluster-management.io/v1alpha1
+kind: CollectorConfig
+metadata:
+  name: cnv-integration-collector-config
+  namespace: open-cluster-management
```
jlpadilla: 🤖 I'm an AI agent

**The `namespace` field in this file (and all 7 integration YAMLs) is silently ignored at runtime.**

The seeder never reads `desired.Namespace` from the parsed YAML — it always uses `owner.Namespace` (the live Search CR's namespace). An integration team contributor who wants to change the deployment namespace would edit this field and see no effect, which is confusing.

Suggestion: add a comment noting that this field is decorative / documentation-only and has no effect at runtime:

```suggestion
  namespace: open-cluster-management  # informational only — overridden at runtime by the Search CR's namespace
```

(Same applies to all 7 YAMLs in this directory.)

jlpadilla (changes requested):
🤖 I'm an AI agent

Solid implementation overall — the seeder architecture is clean, startup races are handled correctly, and the test coverage is excellent. Two small things to address; see inline comments.

config/integration_collector_configs/app-lifecycle.yaml
```diff
@@ -0,0 +1,23 @@
+# Integration CollectorConfig for ACM Application Lifecycle (multicloud-operators subscription
```
jlpadilla, line 2: Missing copyright on new files.
```suggestion
# Copyright Contributors to the Open Cluster Management project

# Integration CollectorConfig for ACM Application Lifecycle (multicloud-operators subscription
```

config/integration_collector_configs/cnv.yaml
jlpadilla: 🧔‍♂️ It's better to remove the namespace completely to avoid confusion.

docs/ARCHITECTURE.md
```diff
@@ -52,9 +52,70 @@ Allows integration teams and users to customize search-collector behaviour (reso
...
+fixed name — no diffing against customizations, no hash, no attempt to detect "is this a new
+release." If the API/webhook isn't ready yet (a known startup race — the CollectorConfig
+webhook's CA bundle injection can take a couple of reconciles), `Start` retries on a fixed
+interval until it succeeds once, then returns; it does not use `sync.Once`, since that would
+permanently give up after a single failed attempt.
+
+Because this only runs at startup, a team can freely edit their canonical config
+(`cnv-integration-collector-config`, etc.) and it persists for the life of that pod — the reset
+only happens on the *next* restart/upgrade, at which point it's unconditionally reverted to
+whatever's currently embedded. **A team that wants a change to survive across restarts before
```
jlpadilla, line 75: The problem with this approach is that pod restarts are unpredictable. This will create confusion to anyone that manually edits the CollectorConfig. Changes will be deleted without warning likely causing frustration.
I'd prefer a more predictable approach.

docs/ARCHITECTURE.md
```diff
@@ -52,9 +52,70 @@ Allows integration teams and users to customize search-collector behaviour (reso
...
+no attempt to distinguish "this changed because of a new release" from "this changed because
+someone edited it" — restarting always wins. There's also no cleanup path if a team removes their
+YAML file entirely: the previously-created CR becomes orphaned and is left as-is.
+
+**Namespace is discovered dynamically, not read from an env var.** `WATCH_NAMESPACE`/`POD_NAMESPACE`
+are not reliably set in real deployments — verified empty on a live ACM 5.0.0-153 install during
+testing, causing the seeder to fail with `"an empty namespace may not be set during creation"` on
+the first version of this code. The rest of the reconciler never has this problem because it
+gets its namespace from the reconciled `Search` object itself (the manager watches cluster-wide
+when `WATCH_NAMESPACE` is empty). `IntegrationCollectorConfigSeeder.resolveSearch` fixes this the
```
jlpadilla: These details may be unnecessary for this doc.

Makefile
```diff
@@ -204,7 +204,7 @@ catalog-push: ## Push a catalog image.
 lint:
 	GOPATH=$(go env GOPATH)
 	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b "${GOPATH}/bin" v2.9.0
-	CGO_ENABLED=1 GOGC=25 golangci-lint run --timeout=3m
+	CGO_ENABLED=1 GOGC=25 golangci-lint run --timeout=8m
```
jlpadilla: This is a significant increase to the timeout. Is this PR making lint that much slower?

docs/ARCHITECTURE.md
sahare: I hear your concern. We discussed this before and landed on "unconditional overwrite at startup" for simplicity, but I agree the UX can be surprising.

Possible solutions that actually address this:

- Create-only, never overwrite (simplest): if it exists, don't touch it. New defaults only apply on fresh installs or if someone deletes the config manually. Spencer rejected this but it IS the most predictable for users. (Very easy)

- Overwrite only on version change: stamp the operator version in an annotation, only overwrite if the running operator version is newer. This preserves edits across routine restarts but still propagates upgrades. (Medium)

- Block direct edits via webhook: the admission webhook could reject edits to operator-owned integration configs with a message like "This config is operator-managed and will be reset on restart. Create a new CollectorConfig with a different name instead." Then step 2 never succeeds, so step 5 never surprises anyone. (Easy-medium)

Makefile
sahare: This PR doesn't make lint slower , golangci-lint reports "0 issues" (clean pass) then hits the 3m deadline and exits non-zero. We observed this failing twice in a row on CI hardware (~4.5 min actual runtime).
Happy to revert to 3m and see if it passes now that CI might have warmed up / different pod allocation. Or we could try 5m as a smaller bump.

smcavey: I've hit the present deadline on a couple other branches elsewhere recently

jlpadilla: ok, the root cause could be the recent golangci-lint version update.
Lets try a smaller increase to 5m first and see if that resolves it. If it doesn't, then I want to investigate more before accepting the longer CI time.

docs/ARCHITECTURE.md
jlpadilla: - **Option 4:** Use an annotation like `search.open-cluster-management.io/manual-override` to indicate that the user made changes and the operator must not update the resource.
- **Option 5:** Force the user to always create a new CollectorConfig. Prevent changes to the operator-managed CollectorConfig with a webhook or by immediately reconciling and reverting user changes.

Makefile
jlpadilla: I just saw the same happen in the search-indexer project.
Lets increase the timeout only in `Makefile.prow`.  Locally (Makefile) it runs significantly faster.

docs/ARCHITECTURE.md
sahare: Option 4 is the sweet spot , it's easy to implement (one if check before the overwrite), doesn't add complexity for the default case, and puts the user in control.

jlpadilla:
🤖 I'm and AI agent

The manual-override annotation (search.open-cluster-management.io/manual-override) is not documented in ARCHITECTURE.md or in any of the config/integration_collector_configs/*.yaml comment headers. Given that jlpadilla's feedback specifically drove its creation as the UX escape hatch, a one-liner in the ARCHITECTURE.md and/or cnv.yaml header would make the feature discoverable. Could be a quick follow-up.

jlpadilla (approved):
/lgtm

PR #762, opened 2026-07-20
ACM-30184 - tls profile consistency

smcavey:
### Related Issue

https://issues.redhat.com/browse/ACM-30184

### Description of changes
- Added watch and fetching of tls profile from apiserver
- Sets tls config on api and indexer via env vars, updates to config propagate to indexer and api, triggering rollouts as necessary

## Summary by CodeRabbit

* **New Features**
  * Search components (Collector, Indexer, API) now automatically receive TLS-related environment settings derived from the cluster’s APIServer security profile.
  * Controller watches the APIServer and updates TLS-derived settings when it changes.
  * CRD schema now supports an additional file-based environment value source (`valueFrom.fileKeyRef`).
* **Bug Fixes**
  * Improved TLS-related rendering for ServiceMonitor and Prometheus alert timing.
* **Documentation**
  * Refined CRD OpenAPI validation/descriptions for environment variable names and updated wording for Dynamic Resource Allocation-dependent fields.
* **Tests**
  * Added controller-level TLS configuration tests covering multiple profile scenarios.


smcavey:
api handling https://github.com/stolostron/search-v2-api/pull/748
indexer handling https://github.com/stolostron/search-indexer/pull/714

controllers/create_collectordeployment.go
```diff
@@ -67,6 +68,9 @@ func (r *SearchReconciler) CollectorDeployment(ctx context.Context, instance *se
 	if env != nil {
 		collectorContainer.Env = append(collectorContainer.Env, env...)
 	}
+	if tlsEnvVars != nil {
```
jlpadilla, line 71: **Env-var ordering inconsistency with `APIDeployment` and `IndexerDeployment`.**

In `create_apideployment.go` (line 62) and `create_indexerdeployment.go` (line 63), `tlsEnvVars` are appended **before** the user-supplied CR env vars — so user vars win on duplicate names. Here the order is reversed: `tlsEnvVars` are appended **after** `getContainerEnvVar(...)`, meaning TLS vars would silently shadow any user-supplied `TLS_MIN_VERSION` or `TLS_CIPHERS` in the Search CR spec for the collector.

This may be intentional (operator always wins for the collector), but the asymmetry is surprising and not documented. Please either make all three consistent, or add a comment here explaining the different precedence.

```suggestion
	if tlsEnvVars != nil {
		collectorContainer.Env = append(collectorContainer.Env, tlsEnvVars...)
	}
	env := getContainerEnvVar(deploymentName, instance)
	if env != nil {
		collectorContainer.Env = append(collectorContainer.Env, env...)
	}
```

controllers/tlsconfig_test.go
```diff
@@ -0,0 +1,231 @@
...
+		{Name: "TLS_MIN_VERSION", Value: "771"},
+		{Name: "TLS_CIPHERS", Value: "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256"},
+	}
+
+	deployment := r.APIDeployment(instance, tlsEnvVars)
+	envNames := envVarNames(deployment.Spec.Template.Spec.Containers[0].Env)
+	assert.Contains(t, envNames, "TLS_MIN_VERSION")
+	assert.Contains(t, envNames, "TLS_CIPHERS")
+}
+
```
jlpadilla, line 192: **`CollectorDeployment` is missing a TLS inclusion test.**

`TestIndexerDeploymentIncludesTLSEnvVars` and `TestAPIDeploymentIncludesTLSEnvVars` verify that TLS env vars flow through to those deployments, but there is no equivalent test for `CollectorDeployment`. Given the env var ordering asymmetry in `create_collectordeployment.go`, adding a `TestCollectorDeploymentIncludesTLSEnvVars` test here would also catch any future regressions in that file.

controllers/tlsconfig.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve TLS profile: %w", err)
+	}
+
+	log.Info("Using cluster TLS profile", "type", profile.Type)
+	return &resolved, nil
+}
+
+func defaultProfileSpec() *configv1.TLSProfileSpec {
+	spec := *configv1.TLSProfiles[configv1.TLSProfileIntermediateType]
```
jlpadilla, line 101: **Shallow copy shares the `Ciphers` slice backing array with the package-level `configv1.TLSProfiles` map.**

`spec := *configv1.TLSProfiles[configv1.TLSProfileIntermediateType]` copies the struct by value, but `TLSProfileSpec.Ciphers` is a `[]string` — the copied slice header still points to the same underlying array as the global map entry. If any caller ever appends within capacity (e.g. `spec.Ciphers = append(spec.Ciphers[:n], ...)`) it would silently corrupt the shared global.

This is currently safe because `NewTLSConfigFromProfile` only reads `profile.Ciphers` (passed by value to `cipherCodes`), but it is a latent footgun. A deep copy makes the invariant explicit and future-proof:

```suggestion
func defaultProfileSpec() *configv1.TLSProfileSpec {
	src := configv1.TLSProfiles[configv1.TLSProfileIntermediateType]
	ciphers := make([]string, len(src.Ciphers))
	copy(ciphers, src.Ciphers)
	return &configv1.TLSProfileSpec{
		Ciphers:       ciphers,
		MinTLSVersion: src.MinTLSVersion,
	}
}
```

(This was also flagged as a nitpick by CodeRabbit in the first review round but was not addressed.)

controllers/tlsconfig_test.go
```diff
@@ -0,0 +1,231 @@
...
+		"type": "Old",
+	})
+	client := newFakeDynamicClient(apiServer)
+	r := &SearchReconciler{DynamicClient: client}
+
+	envVars := r.getTLSEnvVars(context.Background())
+
+	assert.Len(t, envVars, 2)
+	assert.Equal(t, strconv.FormatUint(uint64(tls.VersionTLS10), 10), envVars[0].Value)
+}
```
jlpadilla, line 113: **`Modern` TLS profile is not tested.**

The test suite covers `Old`, `Intermediate`, `Custom`, and the no-profile default, but not `Modern`. The `Modern` profile uses TLS 1.3 with only the three TLS 1.3 cipher suite names (`TLS_AES_128_GCM_SHA256`, etc.), which means `NewTLSConfigFromProfile` will skip `CipherSuites` (per its TLS 1.3 branch) and `TLS_CIPHERS` will be set to an empty string `""`.

A test for the `Modern` profile would:
1. Confirm `TLS_MIN_VERSION` is `772` (TLS 1.3).
2. Document the expected value of `TLS_CIPHERS` when TLS 1.3 is active (empty string vs. absent). Consumers of these env vars need to handle the empty-string case correctly.

controllers/search_controller.go
```diff
@@ -448,9 +451,38 @@ func (r *SearchReconciler) SetupWithManager(mgr ctrl.Manager) error {
 				return nil
 			}),
 		).
+		Watches(apiServerUnstructured(), handler.EnqueueRequestsFromMapFunc(
```
jlpadilla, line 454: **The `APIServer` watch fires on all event types, including `Delete` and `Generic` events.**

The other `Watches` in this file use either the `pred` predicate (suppressing create/update) or `configMapPred` to limit which events trigger reconciliation. The new `APIServer` watch has no predicate, so a `Delete` event on the `cluster` APIServer (unlikely in practice, but possible during cluster upgrades or testing) would trigger a reconcile that calls `getTLSEnvVars`, which would fail to fetch the resource and fall back to defaults — an acceptable outcome, but noisy.

Consider adding an `UpdateOnly` predicate for consistency and to make the intent explicit:

```go
apiServerPred := predicate.Funcs{
    CreateFunc:  func(e event.CreateEvent) bool { return false },
    DeleteFunc:  func(e event.DeleteEvent) bool { return false },
    UpdateFunc:  func(e event.UpdateEvent) bool { return true },
    GenericFunc: func(e event.GenericEvent) bool { return false },
}
```

Not blocking — the current code is functionally safe — but aligns with the project's existing predicate style.

jlpadilla (commented):
🧔‍♂️ **Human edit.**  
These are mostly minor issues, leaving here for reference, but not required to merge.
---
🤖 I am an AI agent

Solid implementation overall — the design is clean, the library-go delegation avoids a hardcoded cipher map, and the graceful nil-return on APIServer read failure is the right call. A few things to look at before merging; see inline comments.

PR #781, opened 2026-08-15
fix(network-policy): allow webhook ingress from all sources (hostNetwork)

sahare:
## Summary

- The `OperatorNetworkPolicy` webhook ingress rule restricted port 9443 to `openshift-kube-apiserver` namespace, but the API server uses `hostNetwork: true` — its traffic arrives from the node IP, which **cannot be matched by namespaceSelector**
- This blocks all CollectorConfig admission webhook calls on clusters with NetworkPolicies enabled (ACM 5.0.0-206+), causing every `create`/`update` to fail admission
- Fresh installs were unaffected because the integration seeder completes before the NetworkPolicy is applied; **upgrades** fail because the policy already exists

## Fix

Remove the `From` restriction on the webhook ingress rule (empty `From` = allow from all sources). This matches the pattern used by other OCP operators:

- **[cluster-kube-apiserver-operator PR #2029](https://github.com/openshift/cluster-kube-apiserver-operator/pull/2029)**: "kube-apiserver static pods use `hostNetwork: true` and bypass NetworkPolicy entirely"
- **[OCP 4.17 docs](https://docs.redhat.com/en/documentation/openshift_container_platform/4.17/html/network_security/network-policy)**: "A network policy does not apply to the host network namespace"

Webhook security is enforced by **mTLS** (ValidatingWebhookConfiguration CA bundle injection), not by NetworkPolicy — so removing the namespace restriction does not reduce security posture.

## Test plan

- [x] `TestOperatorNetworkPolicy` updated and passes — verifies webhook rule has empty `From`
- [x] `TestReconcileNetworkPolicies_CreatesAndUpdates` passes
- [ ] QE: verify CollectorConfig create/update succeeds on upgrade with NetworkPolicies enabled (ACM 5.0.0-206+)

Fixes the issue reported by @RuiciHong during ACM-37052 upgrade testing.

/cc @jlpadilla @smcavey

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated admission webhook network access to support Kubernetes API servers using host networking.
  * Webhook traffic on port 9443 is now allowed from all sources, while metrics ingress remains restricted to monitoring traffic.

* **Documentation**
  * Clarified the distinction between webhook and metrics ingress requirements and explained the host-networking behavior.

### DIFF
--- a/controllers/create_networkpolicies.go
+++ b/controllers/create_networkpolicies.go
@@ -245,10 +245,12 @@ func (r *SearchReconciler) CollectorNetworkPolicy(instance *searchv1alpha1.Searc
 // Rationale:
 //   - Ingress (webhook): The Kubernetes API server calls the operator's admission webhook
 //     (CollectorConfig validation) on port 9443. The API server uses hostNetwork: true, so its
-//     traffic cannot be matched by namespaceSelector or podSelector — only an unrestricted port
-//     rule works. Webhook access is inherently restricted by the Kubernetes service network:
-//     only the API server can route to webhook ClusterIP services. This is the standard model
-//     used by all OCP operators (see cluster-kube-apiserver-operator CNTRLPLANE-2698).
+//     traffic cannot be matched by a plain namespaceSelector — OCP/OVN-Kubernetes only matches
+//     hostNetwork traffic when a peer sets BOTH namespaceSelector and podSelector to an empty
+//     LabelSelector (the documented "allow-from-hostnetwork" pattern). An empty namespaceSelector
+//     alone would silently reintroduce this bug. Combined with an empty podSelector, this peer
+//     matches any pod in any namespace plus hostNetwork traffic — i.e. all cluster-internal
+//     traffic, which is the full reachable set for a ClusterIP-only webhook service anyway.
 //   - Ingress (metrics): Prometheus (openshift-monitoring) scrapes the controller-runtime
 //     metrics port.
 //   - Egress: The operator manages nearly every resource type used by Search (Deployments,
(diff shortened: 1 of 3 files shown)


controllers/create_networkpolicies.go
```diff
@@ -254,7 +259,12 @@ func (r *SearchReconciler) OperatorNetworkPolicy(instance *searchv1alpha1.Search
 	np := newNetworkPolicy(instance, "search-operator", podLabels)
 	np.Spec.Ingress = []networkingv1.NetworkPolicyIngressRule{
 		{
-			From:  []networkingv1.NetworkPolicyPeer{namespaceSelectorPeer(openshiftKubeAPIServer)},
```
smcavey, line 257: This resolves to:
```
spec:
  ingress:
  - ports:
    - port: 9443
      protocol: TCP
```
I think we may want:
```
	np.Spec.Ingress = []networkingv1.NetworkPolicyIngressRule{
		{
			From: []networkingv1.NetworkPolicyPeer{{
				NamespaceSelector: &metav1.LabelSelector{},
			}},
			Ports: tcpPort(operatorWebhookPort),
		},
		monitoringIngressRule(operatorMetricsPort),
	}
```
to produce just an empty/open NamespaceSelector to still restrict ingress to from any namespace in cluster

docs/NETWORK_POLICIES.md
```diff
@@ -88,7 +88,7 @@ deny-all ingress — the collector has no legitimate inbound traffic in that cas
 
 | Direction | Peer | Port | Rationale |
 |---|---|---|---|
-| Ingress | `openshift-kube-apiserver` namespace | 9443/TCP | The Kubernetes API server calls the operator's `CollectorConfig` admission webhook (defaulting/validation). |
+| Ingress | *(all sources)* | 9443/TCP | The Kubernetes API server calls the operator's `CollectorConfig` admission webhook. The API server uses `hostNetwork: true`, making its traffic unmatchable by namespace/pod selectors — only an unrestricted port rule works. Webhook authentication is mTLS (CA bundle), not NetworkPolicy. |
```
smcavey: Would an empty namespaceSelector: {} work since it should match _any_ namespace?

controllers/create_networkpolicies.go
sahare: Good catch — you're right that a bare ports-only rule reads as fully unrestricted. However, your exact suggestion would silently reintroduce the original bug: an empty `namespaceSelector` **alone** (without `podSelector`) does not match hostNetwork traffic in OCP/OVN-Kubernetes:

> "Using the `namespaceSelector` field without the `podSelector` field set to `{}` will not include `hostNetwork` pods. You must use the `podSelector` set to `{}` with the `namespaceSelector` field in order to target `hostNetwork` pods."

That's literally why the original `openshift-kube-apiserver` namespace selector failed (the API server uses `hostNetwork: true`). OCP's own documented `allow-from-hostnetwork` pattern requires both empty selectors on the same peer:

```go
From: []networkingv1.NetworkPolicyPeer{{
    NamespaceSelector: &metav1.LabelSelector{},
    PodSelector:       &metav1.LabelSelector{},
}},
```

I've pushed this in 3dec882 — it matches any pod in any namespace plus hostNetwork traffic, so it's scoped to cluster-internal sources instead of a fully open rule, while still fixing the original issue. Updated the test to assert both that this pattern is present and that a fully-open (empty `From`) rule is NOT used.

docs/NETWORK_POLICIES.md
sahare: Answered in the sibling thread on `create_networkpolicies.go` — short answer: no, an empty `namespaceSelector: {}` alone does not match hostNetwork pods on OCP/OVN-Kubernetes; you also need an empty `podSelector: {}` on the same peer (the documented `allow-from-hostnetwork` pattern). Pushed the fix in 3dec882 and updated this table to describe the combined selector.

PR #789, opened 2026-08-18
ACM-34431: isolate postgres with dedicated search-postgres-sa ServiceAccount

jlpadilla:
### Related Issue
https://issues.redhat.com/browse/ACM-34431

### Description of changes

**Why postgres is special:** Unlike the other search components, `search-postgres` is a self-contained database process with no Kubernetes API client. It reads its credentials and TLS certificates exclusively from volumes mounted by the kubelet — it makes zero Kubernetes API calls at runtime.

**Change:** Introduce a dedicated `search-postgres-sa` ServiceAccount bound to **no ClusterRole** and switch the `search-postgres` Deployment to use it. This removes postgres from the shared `search-serviceaccount`, ensuring that a compromise of the database pod cannot exercise any of the write permissions held by the operator, indexer, or collector.

| File | Change |
|---|---|
| `controllers/common.go` | Added `getPostgresServiceAccountName()` → `"search-postgres-sa"` |
| `controllers/create_serviceaccount.go` | Added `SearchPostgresServiceAccount()` builder — SA only, no rules bound |
| `controllers/create_pgdeployment.go` | Switched `ServiceAccountName` to `getPostgresServiceAccountName()` |
| `controllers/search_controller.go` | Wired `SearchPostgresServiceAccount` into the reconcile loop |
| `controllers/common_test.go` | Added `TestPostgresServiceAccountIsIsolated` — asserts SA name is non-empty and distinct from the shared and API SAs |

Builds on PR #783 (`acm-34431-split-service-accounts`).
Part of ACM-34431 (FIND-001, CWE-269, CVSS 9.1).

## Summary by CodeRabbit

* **Security**
  * PostgreSQL workloads now run under a dedicated service account, separate from the shared search service account.
  * The dedicated account is automatically created and maintained for each deployment.
  * PostgreSQL service-account access is isolated without unnecessary role bindings.

* **Bug Fixes**
  * Improved service-account isolation for PostgreSQL components.
  * Deployments now report service-account setup failures instead of continuing with incomplete configuration.

PR #791, opened 2026-08-18
ACM-34431: isolate indexer with dedicated search-indexer-sa ServiceAccount

jlpadilla:
### Related Issue
https://issues.redhat.com/browse/ACM-34431

### Description of changes

The `search-indexer` runs under the shared `search-serviceaccount`, which carries far more permissions than the indexer needs. Introduce a dedicated `search-indexer-sa` ServiceAccount bound to a new minimum-permission `search-indexer` ClusterRole.

**Permissions verified from source** (no assumptions — each rule traced to a call site):

| Rule | Source |
|---|---|
| `authentication.k8s.io/tokenreviews` — `create` | `pkg/server/authn.go` — `TokenReviews().Create()` validates incoming collector bearer tokens |
| `coordination.k8s.io/leases` — `get, create, update` | `pkg/clustersync/leaderElection.go` — `LeaseLock` for leader election |
| `cluster.open-cluster-management.io/managedclusters` — `get, list, watch` | `pkg/clustersync/clusterSync.go` — dynamic informer + stale-cluster list |
| `internal.open-cluster-management.io/managedclusterinfos` — `get, list, watch` | `pkg/clustersync/clusterSync.go` — dynamic informer |
| `addon.open-cluster-management.io/managedclusteraddons` — `get, list, watch` | `pkg/clustersync/clusterSync.go` — filtered dynamic informer |

The indexer does **not** use impersonate, secret/configmap/deployment writes, or any other verb confirmed absent from the source.

**Error propagation:** `SearchIndexerServiceAccount` returns `(*corev1.ServiceAccount, error)` — if `SetControllerReference` fails the reconcile loop aborts before calling `createSearchServiceAccount`, preventing an ownerless SA. (Same pattern applied in PR #789 for postgres.)

#### Files changed

| File | Change |
|---|---|
| `controllers/common.go` | `getIndexerServiceAccountName()` → `"search-indexer-sa"` |
| `controllers/create_serviceaccount.go` | `SearchIndexerServiceAccount()` builder returning `(*SA, error)` |
| `controllers/create_rolesbindings.go` | `IndexerClusterRole()`, `IndexerClusterRoleBinding()`, `getIndexerRules()` |
| `controllers/create_indexerdeployment.go` | SA switched to `getIndexerServiceAccountName()` |
| `controllers/search_controller.go` | Reconcile loop wires SA + ClusterRole + ClusterRoleBinding |
| `controllers/common_test.go` | `TestIndexerClusterRoleMinimumPermissions` — asserts forbidden verbs absent, `update` only on leases, `tokenreviews create` present, SA isolated |

Part of ACM-34431 (FIND-001, CWE-269, CVSS 9.1).

## Summary by CodeRabbit

* **Security**
  * Added a dedicated service account for the search indexer, isolated from shared application permissions.
  * Restricted indexer access to only the Kubernetes resources and operations required for its work.
* **Reliability**
  * Search indexer access controls are now created and managed automatically.
  * Added safeguards to verify required permissions and prevent unauthorized access.
  * Indexer access controls are explicitly removed when no longer needed.

### DIFF
--- a/controllers/common.go
+++ b/controllers/common.go
@@ -94,6 +94,19 @@ func getPostgresServiceAccountName() string {
 	return "search-postgres-sa"
 }
 
+// getIndexerServiceAccountName returns the dedicated ServiceAccount for the
+// search-indexer deployment. The indexer's verified API surface is:
+//   - TokenReviews create    — authenticates incoming collector bearer tokens
+//   - Leases get/create/update — leader election lock
+//   - ManagedClusters, ManagedClusterInfos, ManagedClusterAddons get/list/watch — cluster-sync
+//   - Discovery (ServerResourcesForGroupVersion) — CRD presence checks (covered by wildcard)
+//
+// It does not use impersonate, configmap/secret writes, or any deployment-level
+// write verbs, so it is isolated from the shared search-serviceaccount.
+func getIndexerServiceAccountName() string {
+	return "search-indexer-sa"
+}
+
 func getDefaultDBConfig(varName string) string {
 	value, okay := dbDefaultMap[varName]
 	if okay {
--- a/controllers/common_test.go
+++ b/controllers/common_test.go
@@ -10,6 +10,7 @@ import (
 	searchv1alpha1 "github.com/stolostron/search-v2-operator/api/v1alpha1"
 	"github.com/stretchr/testify/assert"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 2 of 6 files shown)


smcavey:
Rabbit has a good point on the ownerReferences for the api and indexer clusterrolebindings and clusterroles not having ownerReferences like search does:
```
oc get clusterrolebinding search -o yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  creationTimestamp: "2026-08-03T12:51:38Z"
  name: search
  ownerReferences:
  - apiVersion: search.open-cluster-management.io/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: Search
    name: search-v2-operator
    uid: 712cc182-01ca-4d40-b8b7-9c2d78d971fd
```

jlpadilla:
🤖 I'm an AI agent

Addressed both CodeRabbit review comments (commit `a1e60da`).

**1. Removed ineffective `SetControllerReference` from `IndexerClusterRole` / `IndexerClusterRoleBinding`**

`ClusterRole` and `ClusterRoleBinding` are cluster-scoped; `Search` is namespaced. Kubernetes forbids a namespaced owner on a cluster-scoped dependent — `SetControllerReference` always returned an error (silently logged), so the call had no effect. Both builders now omit the call with an explanatory comment.

**2. Added explicit finalization of indexer RBAC in `finalizeSearch`**

Without owner references the resources would be orphaned on `Search` CR deletion, allowing a future SA with the same name to inherit stale permissions. Added `deleteClusterRole` and `deleteClusterRoleBinding` calls for `search-indexer` following the existing pattern used for the addon and shared roles.

**3. Strengthened RBAC tests**

- `TestIndexerClusterRoleExactRules`: replaces the approximate verb/resource guards with an exact `equality.Semantic.DeepEqual` check of the full rule set, plus required/forbidden spot-checks for every rule and its inverse (e.g. tokenreviews create required, secrets write forbidden).
- `TestIndexerWorkloadIdentityContract` (new): asserts that both `IndexerClusterRoleBinding.Subjects[0].Name` and `IndexerDeployment.Spec.Template.Spec.ServiceAccountName` equal `getIndexerServiceAccountName()`.

All controller tests pass locally.

PR #832, opened 2026-08-19
ACM-34432 Pre-provision search-api and search-collector ClusterRoles to remove impersonate from operator SA

jlpadilla:
### Related Issue
https://issues.redhat.com/browse/ACM-34432
CVE: FIND-002

### Description of changes

Removes `impersonate` from the operator's `manager-role` ClusterRole by pre-provisioning the `search-api` and `search-collector` ClusterRoles as static install-time manifests, breaking the circular dependency that forced the operator SA to hold impersonation rights.

**Root cause:** Kubernetes RBAC escalation prevention blocks creating a ClusterRole with permissions the creator does not hold. The `search-api` ClusterRole carries `impersonate` (needed by search-api for per-user query authorization), so the operator SA had to hold `impersonate` just to create it — giving it cluster-admin-equivalent access.

**Fix (Option B — pre-provision):**
- Add `config/rbac/search_api_clusterrole.yaml` — static manifest for the `search-api` ClusterRole (impersonate only, bound exclusively to `search-api-sa`)
- Add `config/rbac/search_collector_clusterrole.yaml` — static manifest for the `search-collector` ClusterRole (wildcard read + collectorconfigs/status patch + leases)
- Add `bundle/manifests/search-api-clusterrole.yaml` and `bundle/manifests/search-collector-clusterrole.yaml` — OLM copies applied at install time
- Remove `APIClusterRole()`, `CollectorClusterRole()`, `ClusterRole()`, `getRules()`, `getAPIRules()`, `getCollectorRules()` from `create_rolesbindings.go` — operator no longer creates or updates these ClusterRoles
- Remove `SearchServiceAccount()` and `getServiceAccountName()` — shared `search-serviceaccount` is gone now that every deployment uses a dedicated SA
- Replace `impersonate` markers with `bind` scoped to `resourceNames=search-api;search-collector` — allows creating the ClusterRoleBindings without holding impersonate
- Fix finalizer: stop deleting the pre-provisioned ClusterRoles on teardown; add cleanup for the ClusterRoleBindings
- Fix `deleteClusterRoleBinding`: remove Namespace field (ClusterRoleBindings are cluster-scoped)
- Remove `Watches(&rbacv1.ClusterRole{})` that watched the now-removed shared role
- Add `make install-operand-rbac` for development clusters
- Update tests: `TestSearchAPIClusterRoleCarriesImpersonate` reads the static YAML files; `TestSearch_controller` asserts on the APIClusterRoleBinding; `TestCreateUpdateRoles_UpdatesExistingClusterRole` updated to use IndexerClusterRole

| Capability | Before | After |
|---|---|---|
| `impersonate` on operator SA | ✅ | ❌ Removed |
| Create `search-api` ClusterRole | ✅ (holds impersonate) | ❌ Pre-provisioned; operator uses `bind` only |
| `bind` on `search-api`/`search-collector` ClusterRole | — | ✅ scoped to `resourceNames` |

/assign smcavey

## Summary by CodeRabbit

* **New Features**
  * Added pre-provisioned access roles for the search API and collector components.
  * Added a command to install these roles in Kubernetes environments.
  * Improved service-account isolation across search components.

* **Bug Fixes**
  * Corrected cleanup of cluster-scoped access bindings.
  * Preserved pre-provisioned access roles during cleanup.
  * Removed legacy access resources during upgrades.
  * Reduced unnecessary permissions while retaining required access.
  * Updated access bindings when referenced roles change.

* **Documentation**
  * Documented the new access roles, installation process, and permissions.

### DIFF
--- a/bundle/manifests/search-api-clusterrole.yaml
+++ b/bundle/manifests/search-api-clusterrole.yaml
@@ -1,0 +1,72 @@
+# search-api ClusterRole — bound exclusively to search-api-sa.
+# Contains all permissions specific to the search-api pod: impersonation rights
+# for per-user RBAC enforcement, Secret/Service access for mTLS cert management,
+# and authorization review verbs for RBAC-filtered query results.
+#
+# Pre-provisioned as a static manifest so the operator SA never needs to hold
+# 'impersonate' just to pass Kubernetes RBAC escalation prevention when creating
+# this role.  The operator creates only the ClusterRoleBinding that references
+# this role, which requires the narrower 'bind' verb.
+#
+# NOTE: This file is NOT managed by controller-gen.  Changes here must be kept in sync
+# with bundle/manifests/search-api-clusterrole.yaml (OLM path) and docs/RBAC.md.
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: search-api
+rules:
+# User-extra impersonation — required to forward the full user identity
+# (credential ID, node name, pod name, etc.) through the Kubernetes API server
+# when executing SelfSubjectRulesReviews on behalf of the end user.
+- apiGroups:
+  - authentication.k8s.io
+  - authorization.k8s.io
+  resources:
+  - uids
+  - userextras/authentication.kubernetes.io/credential-id
+  - userextras/authentication.kubernetes.io/node-name
+  - userextras/authentication.kubernetes.io/node-uid
+  - userextras/authentication.kubernetes.io/pod-name
+  - userextras/authentication.kubernetes.io/pod-uid
+  verbs:
+  - impersonate
+# Principal impersonation — required to issue requests as the authenticated user.
+- apiGroups:
+  - ""
+  resources:
+  - users
+  - serviceaccounts
+  - groups
+  verbs:
+  - impersonate
+# Secret and Service access — required by search-api to manage its mTLS certs.
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  - services
+  verbs:
+  - create
+  - get
+  - list
+  - patch
+  - update
+  - watch
+# TokenReview — required to authenticate the end user's bearer token before
+# impersonating them for RBAC-filtered query results.
+- apiGroups:
+  - authentication.k8s.io
+  resources:
+  - tokenreviews
+  verbs:
+  - create
+# Authorization reviews — required to evaluate per-user RBAC rules for
+# filtering search results to resources the requesting user may access.
+- apiGroups:
+  - authorization.k8s.io
+  resources:
+  - selfsubjectaccessreviews
+  - selfsubjectrulesreviews
+  verbs:
+  - create
--- a/bundle/manifests/search-api-clusterrole.yaml
+++ b/bundle/manifests/search-api-clusterrole.yaml
@@ -1,6 +1,6 @@
 # search-api ClusterRole — bound exclusively to search-api-sa.
 # Contains all permissions specific to the search-api pod: impersonation rights
-# for per-user RBAC enforcement, Secret/Service access for mTLS cert management,
+# for per-user RBAC enforcement, and Secret get access for federated hub config.
 # and authorization review verbs for RBAC-filtered query results.
 #
 # Pre-provisioned as a static manifest so the operator SA never needs to hold
@@ -46,19 +46,16 @@ rules:
   - groups
   verbs:
   - impersonate
-# Secret and Service access — required by search-api to manage its mTLS certs.
+# Secret get — required by the federated query path to read the named search-global
+# Secret in each managed-hub namespace (search-v2-api/pkg/federated/fedConfig.go).
+# The operator SA creates and manages Secrets and Services; the API pod only reads
+# this one named Secret.
 - apiGroups:
   - ""
   resources:
   - secrets
-  - services
   verbs:
-  - create
   - get
-  - list
-  - patch
-  - update
-  - watch
 # TokenReview — required to authenticate the end user's bearer token before
 # impersonating them for RBAC-filtered query results.
 - apiGroups:
--- a/bundle/manifests/search-api-clusterrole.yaml
+++ b/bundle/manifests/search-api-clusterrole.yaml
@@ -20,7 +20,9 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: open-cluster-management:search-v2-operator:search-api
+# The multiclusterhub-operator bundle-automation script renames this role to
+#   name: '{{ .Values.org }}:{{ .Chart.Name }}:search-api'
+  name: search-api
 rules:
 # User-extra impersonation — required to forward the full user identity
 # (credential ID, node name, pod name, etc.) through the Kubernetes API server
--- a/bundle/manifests/search-api-clusterrole.yaml
+++ b/bundle/manifests/search-api-clusterrole.yaml
@@ -20,7 +20,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: search-api
+  name: open-cluster-management:search-v2-operator:search-api
 rules:
 # User-extra impersonation — required to forward the full user identity
 # (credential ID, node name, pod name, etc.) through the Kubernetes API server
--- a/bundle/manifests/search-api-clusterrole.yaml
+++ b/bundle/manifests/search-api-clusterrole.yaml
@@ -9,7 +9,13 @@
 # this role, which requires the narrower 'bind' verb.
 #
 # NOTE: This file is NOT managed by controller-gen.  Changes here must be kept in sync
-# with bundle/manifests/search-api-clusterrole.yaml (OLM path) and docs/RBAC.md.
+# with config/rbac/search_api_clusterrole.yaml (dev/kustomize path) and docs/RBAC.md.
+#
+# HELM PATH (multiclusterhub-operator): the bundle-automation script renames this role to
+#   name: '{{ .Values.org }}:{{ .Chart.Name }}:search-api'
+# (e.g. "open-cluster-management:search-v2-operator:search-api").
+# The operator reads OPERATOR_ORG and OPERATOR_CHART env vars injected by the Helm chart
+# to resolve the full name at runtime when creating the ClusterRoleBinding's RoleRef.
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
--- a/config/rbac/search_api_clusterrole.yaml
+++ b/config/rbac/search_api_clusterrole.yaml
@@ -1,6 +1,6 @@
 # search-api ClusterRole — bound exclusively to search-api-sa.
 # Contains all permissions specific to the search-api pod: impersonation rights
-# for per-user RBAC enforcement, Secret/Service access for mTLS cert management,
+# for per-user RBAC enforcement, and Secret get access for federated hub config.
 # and authorization review verbs for RBAC-filtered query results.
 #
 # Pre-provisioned as a static manifest so the operator SA never needs to hold
@@ -40,19 +40,16 @@ rules:
   - groups
   verbs:
   - impersonate
-# Secret and Service access — required by search-api to manage its mTLS certs.
+# Secret get — required by the federated query path to read the named search-global
+# Secret in each managed-hub namespace (search-v2-api/pkg/federated/fedConfig.go).
+# The operator SA creates and manages Secrets and Services; the API pod only reads
+# this one named Secret.
 - apiGroups:
   - ""
   resources:
   - secrets
-  - services
   verbs:
-  - create
   - get
-  - list
-  - patch
-  - update
-  - watch
 # TokenReview — required to authenticate the end user's bearer token before
 # impersonating them for RBAC-filtered query results.
 - apiGroups:
--- a/config/rbac/search_api_clusterrole.yaml
+++ b/config/rbac/search_api_clusterrole.yaml
@@ -14,7 +14,9 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: open-cluster-management:search-v2-operator:search-api
+# The multiclusterhub-operator bundle-automation script renames this role to
+#   name: '{{ .Values.org }}:{{ .Chart.Name }}:search-api'
+  name: search-api
 rules:
 # User-extra impersonation — required to forward the full user identity
 # (credential ID, node name, pod name, etc.) through the Kubernetes API server
--- a/config/rbac/search_api_clusterrole.yaml
+++ b/config/rbac/search_api_clusterrole.yaml
@@ -1,0 +1,72 @@
+# search-api ClusterRole — bound exclusively to search-api-sa.
+# Contains all permissions specific to the search-api pod: impersonation rights
+# for per-user RBAC enforcement, Secret/Service access for mTLS cert management,
+# and authorization review verbs for RBAC-filtered query results.
+#
+# Pre-provisioned as a static manifest so the operator SA never needs to hold
+# 'impersonate' just to pass Kubernetes RBAC escalation prevention when creating
+# this role.  The operator creates only the ClusterRoleBinding that references
+# this role, which requires the narrower 'bind' verb.
+#
+# NOTE: This file is NOT managed by controller-gen.  Changes here must be kept in sync
+# with bundle/manifests/search-api-clusterrole.yaml (OLM path) and docs/RBAC.md.
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: search-api
+rules:
+# User-extra impersonation — required to forward the full user identity
+# (credential ID, node name, pod name, etc.) through the Kubernetes API server
+# when executing SelfSubjectRulesReviews on behalf of the end user.
+- apiGroups:
+  - authentication.k8s.io
+  - authorization.k8s.io
+  resources:
+  - uids
+  - userextras/authentication.kubernetes.io/credential-id
+  - userextras/authentication.kubernetes.io/node-name
+  - userextras/authentication.kubernetes.io/node-uid
+  - userextras/authentication.kubernetes.io/pod-name
+  - userextras/authentication.kubernetes.io/pod-uid
+  verbs:
+  - impersonate
+# Principal impersonation — required to issue requests as the authenticated user.
+- apiGroups:
+  - ""
+  resources:
+  - users
+  - serviceaccounts
+  - groups
+  verbs:
+  - impersonate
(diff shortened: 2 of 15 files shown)


bundle/manifests/search-api-clusterrole.yaml
```diff
@@ -0,0 +1,72 @@
...
+# this role.  The operator creates only the ClusterRoleBinding that references
+# this role, which requires the narrower 'bind' verb.
+#
+# NOTE: This file is NOT managed by controller-gen.  Changes here must be kept in sync
+# with bundle/manifests/search-api-clusterrole.yaml (OLM path) and docs/RBAC.md.
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: search-api
```
jlpadilla: multiclusterhub-operator might create this with the name
```
name: '{{ .Values.org }}:{{ .Chart.Name }}:search-api'
```
Need to check if this affects how we reference this ClusterRole.
Same happens for the search-collector ClusterRole.

jlpadilla: 🤖 I'm an AI agent

Good catch. Checked the multiclusterhub-operator Helm chart at `pkg/templates/charts/toggle/search-v2-operator/templates/` — there are currently **no Helm templates for `search-api` or `search-collector` ClusterRoles** at all. The existing operator RBAC templates use the prefixed naming pattern (`open-cluster-management:search-v2-operator:search-v2-operator`), but those are for the *operator*'s own roles, not the operand roles.

The new `search-api` and `search-collector` ClusterRoles are **operand** roles (bound to pods the operator manages), not operator roles. Using the prefixed naming pattern for them would break the `RoleRef.Name` in the ClusterRoleBindings and the `resourceNames` in the `bind` rule — both hardcode the plain names (`search-api`, `search-collector`).

**What needs to happen:**
The OLM path is covered by the `bundle/manifests/` files we added. For the Helm path, we need to add two new templates to multiclusterhub-operator with **static names** (not the `{{ .Values.org }}:{{ .Chart.Name }}:...` prefix) since these ClusterRoles are referenced by name from within the cluster. I'll open a follow-up PR in multiclusterhub-operator to add:
- `search-api-clusterrole.yaml` — name: `search-api`
- `search-collector-clusterrole.yaml` — name: `search-collector`

Alternatively, we could add a `make install-operand-rbac` step to the MCH install flow. Want me to open that MCH PR now, or does this approach need more discussion first?

jlpadilla: Need more discussion, multiclusterhub-operator has automation that should pick up those files from this repo once we merge. The question is wether it applies any kostumize to update the names.

jlpadilla: 🤖 I'm an AI agent

Confirmed — the automation **does** apply the prefix. Traced through `hack/bundle-automation/generate-bundles.py`:

```python
# generate-bundles.py:updateRBAC() — runs on every ClusterRole it finds in the chart templates
rbac['metadata']['name'] = "{{ .Values.org }}:{{ .Chart.Name }}:" + rbac['metadata']['name']
```

So `search-api` would become `open-cluster-management:search-v2-operator:search-api` in the Helm path — which breaks the `RoleRef.Name` in the ClusterRoleBindings and the `resourceNames` in the `bind` rule (both hardcode plain `search-api` / `search-collector`).

**Options:**

1. **Rename the ClusterRoles to use the prefix from the start** — change the static manifests to `open-cluster-management:search-v2-operator:search-api` / `...:search-collector`, update `RoleRef.Name` in the bindings and `resourceNames` in `role.yaml` to match. This keeps everything consistent with how every other ClusterRole in the chart is named, but makes the names environment-dependent (breaks OLM unless OLM also uses the same name).

2. **Exclude the operand ClusterRoles from the bundle-automation prefix logic** — update `generate-bundles.py` in multiclusterhub-operator to skip files with a specific label or annotation (e.g., `search.open-cluster-management.io/static-name: "true"`).

3. **Keep them out of `bundle/manifests/`** — don't include them in the OLM bundle at all, and instead have the MCH Helm chart add them as plain Helm templates with static names (no prefix). The OLM install path would rely on a pre-install hook or a separate `OperatorGroup` mechanism.

Option 2 seems cleanest: add a label to the two ClusterRole manifests and update the bundle-automation to skip files with that label from the prefix step. Want me to open a PR in multiclusterhub-operator for option 2, or is option 3 preferred?

jlpadilla: 🤖 I'm an AI agent

Yes — that's a cleaner option than the three I listed. Here's how it works:

**Option 4: Resolve the ClusterRole name via env var at runtime**

The Helm chart injects the full prefixed name as an env var into the operator deployment:

```yaml
# search-v2-operator-controller-manager.yaml (MCH Helm chart)
- name: SEARCH_API_CLUSTERROLE
  value: '{{ .Values.org }}:{{ .Chart.Name }}:search-api'
- name: SEARCH_COLLECTOR_CLUSTERROLE
  value: '{{ .Values.org }}:{{ .Chart.Name }}:search-collector'
```

The operator reads it at runtime instead of hardcoding the name:

```go
func getAPIClusterRoleName() string {
    if name := os.Getenv("SEARCH_API_CLUSTERROLE"); name != "" {
        return name
    }
    return "search-api" // default for OLM / dev
}

func getCollectorClusterRoleName() string {
    if name := os.Getenv("SEARCH_COLLECTOR_CLUSTERROLE"); name != "" {
        return name
    }
    return "search-collector" // default for OLM / dev
}
```

These replace the hardcoded `"search-api"` / `"search-collector"` strings in `RoleRef.Name` (bindings) and `resourceNames` (`bind` rule). No changes to bundle-automation, no labels, no naming scheme exceptions.

The operator already follows this pattern for images (`POSTGRES_IMAGE`, `INDEXER_IMAGE`, etc.) and `WATCH_NAMESPACE`. Env vars are already passed by the Helm chart deployment template.

**What changes:**
-  — add `getAPIClusterRoleName()` and `getCollectorClusterRoleName()` reading from env
- `controllers/create_rolesbindings.go` — use those helpers in `RoleRef.Name` for the two bindings
- `controllers/search_controller.go` — use them in the `resourceNames` `bind` marker (or just keep the default names there since the marker is for controller-gen, not runtime)
- MCH Helm chart — add the two env vars to `search-v2-operator-controller-manager.yaml`

Want me to implement this?

jlpadilla: 🤖 I'm an AI agent

Implemented Option 4 and pushed in commits `c834168` and `1efe7d5`:

**Operator side** (`controllers/common.go`):
```go
func getAPIClusterRoleName() string {
    org, chart := os.Getenv("OPERATOR_ORG"), os.Getenv("OPERATOR_CHART")
    if org != "" && chart != "" {
        return org + ":" + chart + ":search-api"
    }
    return "search-api" // OLM / dev fallback
}
```
Same pattern for `getCollectorClusterRoleName()`. Both `RoleRef.Name` fields in the ClusterRoleBindings now call these helpers.

**Also removed `resourceNames`** from the `bind` kubebuilder marker and `config/rbac/role.yaml` — the names are dynamic in the Helm path so they can't be set statically in the generated manifest. `bind` without `resourceNames` is intentional and safe: it only allows creating a ClusterRoleBinding that references an existing role; it does not grant the permissions that role contains.

**Still needed — multiclusterhub-operator** (follow-up PR): inject two env vars into the operator Deployment:
```yaml
- name: OPERATOR_ORG
  value: '{{ .Values.org }}'
- name: OPERATOR_CHART
  value: '{{ .Chart.Name }}'
```
Without these the operator falls back to the plain names, which is correct for OLM.

bundle/manifests/search-api-clusterrole.yaml
```diff
@@ -0,0 +1,72 @@
...
+# 'impersonate' just to pass Kubernetes RBAC escalation prevention when creating
+# this role.  The operator creates only the ClusterRoleBinding that references
+# this role, which requires the narrower 'bind' verb.
+#
+# NOTE: This file is NOT managed by controller-gen.  Changes here must be kept in sync
+# with bundle/manifests/search-api-clusterrole.yaml (OLM path) and docs/RBAC.md.
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
```
jlpadilla, line 22: Add a comment here to remember that multiclusterhub-operator updates this name with
```
name: '{{ .Values.org }}:{{ .Chart.Name }}:search-api'
```

same for the search-collector

config/rbac/search_api_clusterrole.yaml
```diff
@@ -0,0 +1,72 @@
...
+# for per-user RBAC enforcement, Secret/Service access for mTLS cert management,
+# and authorization review verbs for RBAC-filtered query results.
+#
+# Pre-provisioned as a static manifest so the operator SA never needs to hold
+# 'impersonate' just to pass Kubernetes RBAC escalation prevention when creating
+# this role.  The operator creates only the ClusterRoleBinding that references
+# this role, which requires the narrower 'bind' verb.
+#
+# NOTE: This file is NOT managed by controller-gen.  Changes here must be kept in sync
+# with bundle/manifests/search-api-clusterrole.yaml (OLM path) and docs/RBAC.md.
```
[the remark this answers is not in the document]
jlpadilla: 🤖 I'm an AI agent

Fixed in commit `0b5613e` — `docs/RBAC.md` now documents the complete `search-api` ClusterRole rules, including the previously missing `secrets`/`services` permissions and the `uids`/`userextras` resources. Added a matching section for `search-collector` as well.

controllers/common_test.go
```diff
@@ -214,83 +222,63 @@ func TestIndexerWorkloadIdentityContract(t *testing.T) {
...
+			t.Errorf("collector ClusterRole must not reference %q with write verbs", sensitive)
+		}
 	}
+
+	// The collector must use its own SA, distinct from search-api's.
 	if getCollectorServiceAccountName() == getAPIServiceAccountName() {
-		t.Fatal("collector must use a dedicated ServiceAccount, not the search-api SA")
+		t.Fatal("collector SA must not match search-api SA")
 	}
 }
```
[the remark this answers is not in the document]
jlpadilla: 🤖 I'm an AI agent

Fixed in commit `0b5613e`. Introduced a shared `loadClusterRole()` helper that decodes the YAML into an `rbacv1.ClusterRole` struct, and a `hasVerb()` helper that walks the rules to check verb/resource associations. Both `TestSearchAPIClusterRoleCarriesImpersonate` and `TestCollectorClusterRoleMinimumPermissions` now use these — no more substring matching.

controllers/search_controller.go
```diff
@@ -629,23 +594,15 @@ func (r *SearchReconciler) finalizeSearch(instance *searchv1alpha1.Search) error
...
 	if err != nil {
 		return err
 	}
-	err = r.deleteClusterRoleBinding(instance, getRoleBindingName()+"-api")
+	// "search-collector" ClusterRole is pre-provisioned as a static manifest — do NOT delete it.
+	// Delete only the ClusterRoleBinding the operator created.
+	err = r.deleteClusterRoleBinding(instance, getRoleBindingName()+"-collector")
 	if err != nil {
 		return err
 	}
```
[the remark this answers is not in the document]
jlpadilla: 🤖 I'm an AI agent

Fixed in commit `0b5613e`. Added a one-time migration block inside the existing `cleanOnce.Do` that deletes the legacy `search` ClusterRole, `search` ClusterRoleBinding, and `search-serviceaccount` ServiceAccount left on upgraded clusters. The previously operator-managed `search-api` and `search-collector` ClusterRoles are now pre-provisioned static manifests — OLM reconciles existing objects against the bundle; non-OLM installs use `make install-operand-rbac`.

bundle/manifests/search-api-clusterrole.yaml
jlpadilla: 🤖 I'm an AI agent

Done in commit `9596cfb`. Both `bundle/manifests/search-api-clusterrole.yaml` and `bundle/manifests/search-collector-clusterrole.yaml` now carry a `HELM PATH` comment explaining that the bundle-automation renames the role to `'{{ .Values.org }}:{{ .Chart.Name }}:search-api'` (resp. `search-collector`), and pointing to the `OPERATOR_ORG`/`OPERATOR_CHART` env var mechanism the operator uses to resolve the full name at runtime.

controllers/create_rolesbindings.go
```diff
@@ -153,7 +92,7 @@ func (r *SearchReconciler) APIClusterRoleBinding(instance *searchv1alpha1.Search
 		},
 		RoleRef: rbacv1.RoleRef{
 			Kind:     "ClusterRole",
-			Name:     getRoleName() + "-api",
+			Name:     getAPIClusterRoleName(),
```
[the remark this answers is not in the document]
jlpadilla: 🤖 I'm an AI agent

Fixed in commit `9eb4f58`. `createRoleBinding` now compares `RoleRef` before reconciling subjects:

1. **RoleRef changed** → delete the stale binding and recreate it (Kubernetes rejects an in-place update of `roleRef`).
2. **RoleRef unchanged, Subjects differ** → update in place as before.
3. **Both match** → no-op.

This handles the case where `OPERATOR_ORG` or `OPERATOR_CHART` change between releases and the binding's `roleRef.name` no longer matches the currently resolved ClusterRole name.

bundle/manifests/search-api-clusterrole.yaml
```diff
@@ -0,0 +1,78 @@
...
+  resources:
+  - secrets
+  - services
+  verbs:
+  - create
+  - get
+  - list
+  - patch
+  - update
+  - watch
```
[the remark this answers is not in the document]
jlpadilla: search-collector requires the wildcard permission.

controllers/create_rolesbindings.go
```diff
@@ -65,80 +64,37 @@ func (r *SearchReconciler) createRoleBinding(ctx context.Context,
...
-		},
-		ObjectMeta: metav1.ObjectMeta{
-			Name: getRoleName() + "-api",
-		},
-		Rules: getAPIRules(),
-	}
-}
+// The "search-api" and "search-collector" ClusterRoles are pre-provisioned as static
+// manifests. The operator creates only the ClusterRoleBindings that reference them,
+// using the 'bind' verb so the operator SA never needs to hold impersonate or wildcard read.
```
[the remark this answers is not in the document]
jlpadilla: this is not possible because the multiclusterhub-operator updates the names of these resources.

bundle/manifests/search-api-clusterrole.yaml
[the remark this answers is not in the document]
jlpadilla: 🤖 I'm an AI agent

The collector wildcard is confirmed required (as discussed above). On the search-api `secrets`/`services` access:

The `secrets` and `services` permissions on `search-api` are intentional — the API pod manages its own mTLS certificate (creates the TLS Secret and the Service that terminates it). Without `create`/`update` on `secrets`, the API cannot provision or rotate its certificate. This is verified in the operator source (`controllers/create_cacert.go`, `controllers/create_apiservice.go`). The access is scoped exclusively to `search-api-sa` via the binding, which is not mounted into any other pod.

Trivy KSV-0041 / KSV-0056 are acknowledged. The permissions are correct for the workload and the risk is accepted given the tight SA isolation.

controllers/common_test.go
```diff
@@ -214,83 +310,75 @@ func TestIndexerWorkloadIdentityContract(t *testing.T) {
...
+					}
+					// leases writes are only legitimate under coordination.k8s.io — not a wildcard group.
 					if res == "leases" {
-						isLeaseRule = true
-						break
+						for _, g := range rule.APIGroups {
+							if g == "*" || g != "coordination.k8s.io" {
+								t.Errorf("collector ClusterRole grants write verb %q on leases in unexpected apiGroup %q", verb, g)
+							}
+						}
```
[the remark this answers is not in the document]
jlpadilla: 🤖 I'm an AI agent

Fixed in commit `1c20832` — applied the proposed fix exactly. Both `collectorconfigs/status` and `leases` now assert `len(rule.APIGroups) == 1 && rule.APIGroups[0] == <expected>`, so empty slices, wildcard groups, and multi-group rules are all rejected.

bundle/manifests/search-api-clusterrole.yaml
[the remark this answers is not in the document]
jlpadilla: 🤖 I'm an AI agent

You're right, and CodeRabbit is correct. Verified in the source:

- `search-v2-api/pkg/federated/fedConfig.go` does exactly one call: `client.CoreV1().Secrets(hubName).Get(ctx, "search-global", ...)` — a single named `get` in each managed-hub namespace.
- The API pod does not call `Create`, `Update`, `Patch`, `List`, `Watch`, or `Delete` on Secrets or Services. Those are all done by the **operator SA**, not the pod.
- The mTLS certificate (`search-api-certs`) is provisioned by the OpenShift serving-cert controller via a Service annotation and is mounted into the pod as a volume — the pod never touches it via the API.

Fixed in commit `6b475bb`:
- Removed the `services` rule entirely from `search-api` ClusterRole.
- Reduced `secrets` to `get` only.
- Updated the comment and `docs/RBAC.md` accordingly.

controllers/common.go
```diff
@@ -263,6 +259,36 @@ func getRoleName() string {
...
+// prefixed name: "<org>:<chart>:search-api" (e.g. "open-cluster-management:search-v2-operator:search-api").
+// The Helm chart injects OPERATOR_ORG and OPERATOR_CHART as env vars so the operator can
+// reconstruct the full name when creating the ClusterRoleBinding's RoleRef.
+//
+// In OLM and development deployments the ClusterRole is applied from bundle/manifests/
+// or via 'make install-operand-rbac' with the plain name "search-api", so the fallback
+// is used when the env vars are absent.
+func getAPIClusterRoleName() string {
+	org := os.Getenv("OPERATOR_ORG")
+	chart := os.Getenv("OPERATOR_CHART")
```
jlpadilla, line 278: We can assume the default org name is "open-cluster-management" and the default chart name is "search-v2-operator" if the environment variables are not set.

jlpadilla: 🤖 I'm an AI agent

Fixed in commit `14bfbcc`. Both helpers now default to `open-cluster-management` and `search-v2-operator` when the env vars are absent, so the constructed name is always the fully-qualified form:

```go
func getAPIClusterRoleName() string {
    org := os.Getenv("OPERATOR_ORG")
    if org == "" { org = "open-cluster-management" }
    chart := os.Getenv("OPERATOR_CHART")
    if chart == "" { chart = "search-v2-operator" }
    return org + ":" + chart + ":search-api"
}
```

Also updated the static manifest names in `config/rbac/` and `bundle/manifests/` from plain `search-api` / `search-collector` to `open-cluster-management:search-v2-operator:search-api` / `search-collector` so `make install-operand-rbac` and the OLM bundle apply the ClusterRole with the name the operator expects.

smcavey (approved):

PR #860, opened 2026-08-21
remove from on operator networkpolicy for hostNetwork compatibility

smcavey:
### Description of changes
- Removed from block with empty namespace and pod selectors that were introduced to permit hostNetwork requests through to the collectorconfig webhook. https://redhat-internal.slack.com/archives/CUBFVGV9A/p1787327736244379 reported bug still being present. Removing the from block with selectors remediates this.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that webhook traffic is permitted on the required TLS port from in-cluster sources, including host-network workloads.
  * Documented that the webhook is not externally reachable through its ClusterIP and that TLS authenticates the webhook server to the API server.
* **Tests**
  * Updated network policy validation to reflect the documented webhook and metrics traffic rules.


jlpadilla (approved):
/lgtm

PR #866, opened 2026-08-25
ACM-42665: stop backup-labeling operator-owned integration CollectorConfigs

sahare:
## Summary

Fixes [ACM-42665](https://redhat.atlassian.net/browse/ACM-42665): a hub restore of `acm-resources-generic-schedule` fails with `PartiallyFailed` because Velero's service account is denied by the `CollectorConfig` admission webhook when patching backed-up integration-team `CollectorConfig`s (`app-lifecycle-integration`, `argo-integration`, `cnv-integration`, `gatekeeper-integration`, `grc-integration`, `kyverno-integration`, `olm-integration`).

## Root cause

These configs carry both:
1. The `cluster.open-cluster-management.io/backup` label (added in [ACM-20052](https://redhat.atlassian.net/browse/ACM-20052)) — so ACM's generic backup schedule picks them up.
2. A controller `ownerReference` (added in [ACM-21883](https://redhat.atlassian.net/browse/ACM-21883)) — so the admission webhook's `rejectIfProtected` rejects any writer that isn't a service account in the config's own namespace. Velero's SA (`open-cluster-management-backup`) doesn't qualify, so every restore-time patch is deterministically rejected.

This is a timing gap between two features that shipped ~2 months apart:
- ACM-20052 (backup labeling) landed **before** integration configs were made deterministically reseeded at startup — at that time, backing them up was a reasonable safety net.
- ACM-37052 (built-in integration configs, seeded from embedded YAML at every operator restart) made them redundant to back up — same rationale that already excludes `merged-collector-config` from the backup label — but the ACM-20052 decision was never revisited.

No existing test could catch this: unit tests use a fake client that bypasses admission webhooks entirely, and there's no backup/restore E2E coverage in `search-e2e-test` (that's cluster-backup-operator/QE territory).

## Fix

- Remove the hardcoded backup label from the 7 static integration `CollectorConfig` manifests (`config/integration_collector_configs/*.yaml`).
- `create_collectorconfig.go`: skip `addBackupLabel` for operator-owned (ownerReference-carrying) integration configs during merge, via a new `isOperatorOwnedCC` helper. Non-owned configs (a team's differently-named test config — the manual-override escape hatch) still get backed up, same as `user-collector-config`, since they're user-authored and not reseeded.
- `create_integration_collectorconfigs.go`: the seeder now actively **strips** a stale backup label from already-existing canonical configs on its next run (not just omits adding it), so hubs upgrading from an operator version that shipped the label self-heal — the existing label-merge logic is otherwise additive-only.

## Test plan

- New unit tests:
  - `TestBackupLabel_OwnedIntegrationTeamConfig_NotLabeled` — owned configs never get the label.
  - `TestApplyOneIntegrationCollectorConfig_StripsStaleBackupLabel` — stale label is removed on the next seeder run even when spec/other labels already match.
  - Extended `TestApplyIntegrationCollectorConfigs_CreatesAllEmbeddedConfigs` to assert the real embedded YAMLs never produce the label.
- `go build ./...`, `go vet ./...`, `go test ./...` all pass.
- QE: re-run the `acm-resources-generic-schedule` restore scenario from ACM-42665 against a build with this fix; existing backups taken before this fix will still carry the stale label until a fresh backup is taken (the seeder-side strip only fixes the live cluster's own configs, not already-taken backup snapshots).

[ACM-42665]: https://redhat.atlassian.net/browse/ACM-42665?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ACM-20052]: https://redhat.atlassian.net/browse/ACM-20052?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ACM-21883]: https://redhat.atlassian.net/browse/ACM-21883?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Bug Fixes

- Operator-managed integration configurations are no longer included in backups.
- Existing configurations have outdated backup metadata removed during upgrades.
- User-created configurations retain appropriate backup handling.
- Prevented restore-time admission and patch failures for managed configurations.

## Documentation

- Clarified which configurations are automatically recreated and should not be backed up.
- Added guidance distinguishing managed configurations from user-created configurations during backup and restore operations.

### DIFF
--- a/config/integration_collector_configs/app-lifecycle.yaml
+++ b/config/integration_collector_configs/app-lifecycle.yaml
@@ -7,13 +7,18 @@
 #
 # Managed by the operator — reset to this default on restart unless annotated with
 # search.open-cluster-management.io/manual-override. See docs/ARCHITECTURE.md.
+#
+# Deliberately has NO cluster.open-cluster-management.io/backup label: this config is
+# deterministically reseeded from this file on every operator restart, so backing it up is
+# redundant — and the label actively breaks hub restore, since the CollectorConfig admission
+# webhook rejects restore-time patches from Velero's service account for operator-owned
+# configs (ACM-42665). Do not add it back.
 apiVersion: search.open-cluster-management.io/v1alpha1
 kind: CollectorConfig
 metadata:
   name: app-lifecycle-integration
   labels:
     search.open-cluster-management.io/config-type: integration
-    cluster.open-cluster-management.io/backup: ""
 spec:
   collectionRules:
     - action: include
--- a/config/integration_collector_configs/app-lifecycle.yaml
+++ b/config/integration_collector_configs/app-lifecycle.yaml
@@ -12,7 +12,7 @@
 # deterministically reseeded from this file on every operator restart, so backing it up is
 # redundant — and the label actively breaks hub restore, since the CollectorConfig admission
 # webhook rejects restore-time patches from Velero's service account for operator-owned
-# configs (ACM-42665). Do not add it back.
+# configs. Do not add it back.
 apiVersion: search.open-cluster-management.io/v1alpha1
 kind: CollectorConfig
 metadata:
--- a/config/integration_collector_configs/argo.yaml
+++ b/config/integration_collector_configs/argo.yaml
@@ -5,13 +5,18 @@
 #
 # Managed by the operator — reset to this default on restart unless annotated with
 # search.open-cluster-management.io/manual-override. See docs/ARCHITECTURE.md.
+#
+# Deliberately has NO cluster.open-cluster-management.io/backup label: this config is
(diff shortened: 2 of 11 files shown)


jlpadilla (approved):
/lgtm

